Issue #14, opened 2017-09-21
RegistriesImpl: List..Async() methods not implemented


lenala:
In RegistriesImpl 4 List...Async() methods not implemented:
```
protected override Task<IPage<Models.RegistryInner>> ListInnerAsync(CancellationToken cancellationToken)
        {
            throw new NotImplementedException();
        }

        protected override Task<IPage<Models.RegistryInner>> ListInnerNextAsync(string nextLink, CancellationToken cancellationToken)
        {
            throw new NotImplementedException();
        }

        protected override Task<IPage<Models.RegistryInner>> ListInnerByGroupAsync(string groupName, CancellationToken cancellationToken)
        {
            throw new NotImplementedException();
        }

        protected async override Task<IPage<Models.RegistryInner>> ListInnerByGroupNextAsync(string nextLink, CancellationToken cancellationToken)
        {
            return await Task.FromResult<IPage<Models.RegistryInner>>(null);
        }
```

Please note: same for ContainerServicesImpl.

PR #21, opened 2017-09-25
Fixes for ACR and ACS list methods

milismsft:


hovsepm:
@milismsft if this is a fix for https://github.com/Azure/azure-libraries-for-net/issues/14 then please close the issue :)

PR #30, opened 2017-10-03
MSI credentials support

anuchandy:


src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
```diff
@@ -80,6 +88,12 @@ public AzureCredentials WithDefaultSubscription(string subscriptionId)
 
         public async override Task ProcessHttpRequestAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
+            if (this.msiTokenProvider != null)
+            {
+                request.Headers.Authorization = await this.msiTokenProvider.GetAuthenticationHeaderAsync(cancellationToken);
```
anuchandy: Credentials won't be cached explicitly. The token provider (MSI extension) running on VM will be responsible for caching the auth token.

hovsepm:
@anuchandy We need to finalize release 1.3 before merging this.

PR #64, opened 2017-10-17
Fix odata query for tags

karataliu:
String value should be single quoted.

Otherwise for '.ListByTag("a","b")', it will report
```
Unhandled Exception: Microsoft.Rest.Azure.CloudException: Invalid $filter 'tagname eq a and tagvalue eq b' specified in the query string.
```

I can get expected result with '.ListByTag("'a'","'b'")


hovsepm:
@karataliu thanks for the PR! Please add test case.

hovsepm:
@karataliu please review and merge PR https://github.com/karataliu/azure-libraries-for-net/pull/1 
It will add test coverage for this PR.

PR #66, opened 2017-10-18
App gateway backend health check support, GA of load balancer support

martinsawicki:
ported from Java PR https://github.com/Azure/azure-sdk-for-java/pull/1945

### DIFF
--- a/src/ResourceManagement/Network/NicIPConfigurationImpl.cs
+++ b/src/ResourceManagement/Network/NicIPConfigurationImpl.cs
@@ -212,17 +212,21 @@ namespace Microsoft.Azure.Management.Network.Fluent
         {
             if (loadBalancer != null)
             {
-                foreach (var pool in loadBalancer.Inner.BackendAddressPools)
+                var pool = loadBalancer.Inner.BackendAddressPools.FirstOrDefault(p => p.Name.Equals(backendName, StringComparison.OrdinalIgnoreCase));
+                if (pool != null)
                 {
-                    if (pool.Name.Equals(backendName, StringComparison.OrdinalIgnoreCase))
-                    {
-                        EnsureLoadBalancerBackendAddressPools().Add(pool);
-                        return this;
-                    }
+                    EnsureLoadBalancerBackendAddressPools().Add(pool);
+                    return this;
+                }
+                else
+                {
+                    throw new InvalidOperationException("Load balancer backend address pool not found.");
                 }
             }
-
-            return null;
+            else
+            {
+                throw new InvalidOperationException("Missing load balancer reference.");
+            }
         }
 
 
@@ -231,18 +235,21 @@ namespace Microsoft.Azure.Management.Network.Fluent
         {
             if (loadBalancer != null)
             {
-                foreach (var rule in loadBalancer.Inner.InboundNatRules)
+                var rule = loadBalancer.Inner.InboundNatRules.FirstOrDefault(r => r.Name.Equals(inboundNatRuleName, StringComparison.OrdinalIgnoreCase));
+                if (rule != null)
                 {
-                    if (rule.Name.Equals(inboundNatRuleName, StringComparison.OrdinalIgnoreCase))
-                    {
-                        EnsureInboundNatRules().Add(rule);
-                        return this;
-                    }
+                    EnsureInboundNatRules().Add(rule);
+                    return this;
+                } 
+                else
+                {
+                    throw new InvalidOperationException("Load balancer inbound NAT rule not found.");
                 }
-
             }
-
-            return null;
+            else
+            {
+                throw new InvalidOperationException("Missing load balancer reference.");
+            }
         }
 
         
@@ -435,8 +442,15 @@ namespace Microsoft.Azure.Management.Network.Fluent
                     EnsureAppGatewayBackendAddressPools().Add(pool);
                     return this;
                 }
+                else
+                {
+                    throw new InvalidOperationException("Application gateway backend not found.");
+                }
             }
-            return null;
+            else
+            {
+                throw new InvalidOperationException("Missing application gateway reference.");
+            }
         }
 
         ///GENMHASH:96F361243DC52D09D92B82A4480228B7:03A275DAD040D9C179FC6ACCCB43DE36
--- a/src/ResourceManagement/Network/NicIPConfigurationImpl.cs
+++ b/src/ResourceManagement/Network/NicIPConfigurationImpl.cs
@@ -11,11 +11,12 @@ namespace Microsoft.Azure.Management.Network.Fluent
     using ResourceManager.Fluent.Core.ResourceActions;
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// Implementation for NicIPConfiguration and its create and update interfaces.
     /// </summary>
-    
+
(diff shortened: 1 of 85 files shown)


src/ResourceManagement/ResourceManager/Domain/IResourceGroups.cs
```diff
@@ -26,14 +27,29 @@ public interface IResourceGroups  :
...
-        Task<bool> CheckExistenceAsync(string name, CancellationToken cancellationToken = default(CancellationToken)
-);
+        bool Contain(string name);
+
+        /// <summary>
+        /// Checks whether resource group exists.
+        /// </summary>
+        /// <param name="name">name The name of the resource group to check. The name is case insensitive</param>
+        /// <returns>true if the resource group exists; false otherwise</returns>
+        [Obsolete("User ContgainAsync() instead.")]
```
hovsepm: >ContgainAsync [](start = 24, length = 13)

typo should be ContainAsync

src/ResourceManagement/ResourceManager/Domain/IResourceGroups.cs
```diff
@@ -26,14 +27,29 @@ public interface IResourceGroups  :
...
+        Task<bool> CheckExistenceAsync(string name, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Checks whether resource group exists.
+        /// </summary>
+        /// <param name="name">name The name of the resource group to check. The name is case insensitive</param>
+        /// <returns>true if the resource group exists; false otherwise</returns>
+        Task<bool> ContainAsync(string name, CancellationToken cancellationToken = default(CancellationToken));
```
hovsepm: >ContainAsync [](start = 19, length = 12)

LINQ analogy suggests plural - Contains. [https://msdn.microsoft.com/en-us/library/bhkz42b3(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/bhkz42b3(v=vs.110).aspx)

hovsepm: i.e. instead of saying "Resource groups contain" I say "Resource group list contains". just saying :) 

---
In reply to: [145505367](https://github.com/Azure/azure-libraries-for-net/pull/66#discussion_r145505367) [](ancestors = 145505367)

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
```diff
@@ -166,89 +167,82 @@ internal NicIPConfigurationImpl WithNewPublicIPAddress()
...
+                    if (pool.Name.Equals(backendName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        EnsureLoadBalancerBackendAddressPools().Add(pool);
+                        return this;
+                    }
                 }
             }
 
-            return this;
+            return null;
```
hovsepm: Returning null here will break Fluent flow, right?

src/ResourceManagement/ResourceManager/Domain/IResourceGroups.cs
martinsawicki: thanks for catching

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
```diff
@@ -166,89 +167,82 @@ internal NicIPConfigurationImpl WithNewPublicIPAddress()
...
-        {
-            var poolRefs = Inner.LoadBalancerBackendAddressPools;
-            if (poolRefs == null)
-            {
-                poolRefs = new List<BackendAddressPoolInner>();
-                Inner.LoadBalancerBackendAddressPools = poolRefs;
             }
 
-            return poolRefs;
+            return null;
```
hovsepm: >return null; [](start = 12, length = 12)

here as well. Shouldn't you return this not to break fluent flow?

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
```diff
@@ -415,5 +409,42 @@ NetworkInterface.Update.IUpdate ISettable<NetworkInterface.Update.IUpdate>.Paren
...
+            if (appGateway != null)
+            {
+                var pool = appGateway.Inner.BackendAddressPools.FirstOrDefault(o => o.Name.Equals(backendName, StringComparison.OrdinalIgnoreCase));
+                if (pool != null)
+                {
+                    EnsureAppGatewayBackendAddressPools().Add(pool);
+                    return this;
+                }
+            }
+            return null;
```
hovsepm: >return null; [](start = 12, length = 12)

Please double-check  returning null in the fluent flow.

src/ResourceManagement/ResourceManager/Domain/IResourceGroups.cs
martinsawicki: I've been debating this myself, but just looking at the context where this will be used:
`resourceGroups().contains()` just sounds awkward (not grammatically correct).
It's another thing when we talk about a String (`myString.Contains("foo")`) which sounds perfect. LINQ had to have general purpose naming, hanging off `IEnumerable`, so I can see why the singular form was preferable there.  
But doesn't it just feel icky where the only place one will be using this method is an object whose name is clearly plural (either as the accessor `.resourceGroups()` or the type `ResourceGroups`) and we'd still be using the singular form?

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: well, so that's an intentional change in the logic I made, because the call here is `WithExisting*()` so if it doesn't exist, you shouldn't be able to proceed.  Would throwing an exception (like InvalidOperationException?) be better you think?

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: actually, same thing here - the flow shouldn't really proceed because it is assuming a child that doesn't exist. 
So I used Java's semantic for missing Map elements: return null rather than throw exception.
(Although .NET Dictionary doesn't quite work that way (unfortunately), HybridDictionary does, so that precedent made me feel it's ok on the .NET side to do the same.Leave a comment

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: yes, this is intentional - same rationale as explained earlier

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
hovsepm: yes. NullRef exceptions are really not giving any useful info.

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
hovsepm: Please throw InvalidOperationException instead.

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
hovsepm: please throw instead of returning null.

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: ok

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: ok

src/ResourceManagement/Network/NicIPConfigurationImpl.cs
martinsawicki: ok

PR #96, opened 2017-11-16
Add implementation for AKS

milismsft:
Beta compat breaks:

 - move ContainerService from under compute resource manager to its own resource manager
 - rename dnsLabel() to dnsPrefix()
 - moved withDiagnostics() to the optional methods
 - removed some agent pool respective properties; these can be accessed now via the AgentPools()
 - unless specified default master pool virtual machine size to "Standard_D2_v2"


hovsepm:
>     <PackageReference Include="Microsoft.Azure.Management.Dns.Fluent" Version="1.3.0" />

you need to add container service dependency here as well. 

---
Refers to: src/ResourceManagement/Azure.Fluent/Microsoft.Azure.Management.Fluent.csproj:40 in e313696. [](commit_id = e3136961b2d563e6b54bf68486105ef7f290c016, deletion_comment = False)

hovsepm:
@milismsft I'll address the only comment regarding rollup project file in my PR.

PR #143, opened 2018-01-25
Enable User Assigned Identity in MSITokenProvider

anuchandy:
Current MSITokenProvider supports only System Assigned MSI, updating provider to support User Assigned Identity MSI as well

### DIFF
--- a/src/ResourceManagement/ResourceManager/Authentication/MSITokenProvider.cs
+++ b/src/ResourceManagement/ResourceManager/Authentication/MSITokenProvider.cs
@@ -4,6 +4,7 @@
 using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
 using Microsoft.Rest;
 using Newtonsoft.Json;
+using System;
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Net.Http.Headers;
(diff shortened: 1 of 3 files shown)


src/ResourceManagement/ResourceManager/Authentication/MSITokenProvider.cs
```diff
@@ -17,27 +18,44 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Authentication
...
+            }
+            else if (this.msiLoginInformation.UserAssignedIdentityClientId != null)
+            {
+                parameters.Add("client_id", this.msiLoginInformation.UserAssignedIdentityClientId);
+            }
+            else if (this.msiLoginInformation.UserAssignedIdentityResourceId != null)
+            {
+                parameters.Add("msi_res_id", this.msiLoginInformation.UserAssignedIdentityResourceId);
+            }
+
```
lenala: Is it correct that only one parameter can be specified?

anuchandy: @lenala these are mutually exclusive so we need if..elseif..elseif..

anuchandy:
@lenala we will be adding a test once the lib is published to nuget, because ignorer to test this I need to run consumer application in an Azure VM, so we need packages in nuget so that app can pull it from the VM.

PR #148, opened 2018-01-30
Added Monitor service.

hovsepm:


src/ResourceManagement/Monitor/ActivityLogsImpl.cs
```diff
@@ -115,14 +115,14 @@ public ActivityLogsImpl WithAllPropertiesInResponse()
         ///GENMHASH:F53D539E431A5CEE4CC1B076D4C77610:BFADD6FEAFCB5CD377ED3E9C10E4C678
         public ActivityLogsImpl FilterByResource(string resourceId)
         {
-            this.filterString = String.Format(" and resourceUri eq '%s'", resourceId);
+            this.filterString = String.Format(" and resourceUri eq '{0}'", resourceId);
```
anuchandy: Instead of String.Format we could use

```C#
this.filterString = $" and resourceUri eq '{resourceId}'"
```

src/ResourceManagement/Monitor/Domain/IMetricDefinition.cs
```diff
@@ -54,7 +54,7 @@ public interface IMetricDefinition :
         /// <summary>
         /// Gets the stage of start time filter definition.
         /// </summary>
-        Microsoft.Azure.Management.Monitor.Fluent.IWithMetricStartTimeFilter DefineQuery { get; }
+        Microsoft.Azure.Management.Monitor.Fluent.IWithMetricStartTimeFilter DefineQuery();
```
anuchandy: Remember to annotate this method in Java with `@Method`

PR #153, opened 2018-01-31
Support MSI for VM & App service to other endpoints 

jianghaolu:


src/ResourceManagement/ResourceManager/Authentication/MSITokenProvider.cs
```diff
@@ -18,13 +18,12 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Authentication
     /// </summary>
     public class MSITokenProvider : ITokenProvider, IBeta
     {
-        private readonly string resource;
+        private string resource;
         private readonly MSILoginInformation msiLoginInformation;
```
anuchandy: jfyi: It seems we're initializing `this.msiLoginInformation` twice in the ctr (which I added by mistake when authored MSITokenProvider initially).

src/ResourceManagement/ResourceManager/Authentication/MSITokenProvider.cs
```diff
@@ -18,13 +18,12 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Authentication
     /// </summary>
     public class MSITokenProvider : ITokenProvider, IBeta
     {
-        private readonly string resource;
+        private string resource;
```
anuchandy: why we removed `readonly`?

PR #162, opened 2018-02-05
Gulpfile fixes

RikkiGibson:
Fixes issues in .NET code generation noted by @milismsft.

At some point a new generator was published which adds the `/Generated` path component for us, so we need to stop adding it ourselves, and the README.md we are consuming from azure-rest-api-specs is starting to contain `clear-output-folder: true`, which we don't want because we have our own method of clearing out generated files in gulpfile.js.

### DIFF
--- a/src/ResourceManagement/gulpfile.js
+++ b/src/ResourceManagement/gulpfile.js
@@ -82,7 +82,7 @@ var handleInput = function (projects, cb) {
 
 var codegen = function (project, cb) {
     const regenManager = args['regenerate-manager'] ? ' --regenerate-manager=true ' : '';
-    var outputDir = path.resolve(mappings[project].dir) + '/Generated';
+    var outputDir = path.resolve(mappings[project].dir);
     if (!args['preserve']) {
         deleteFolderRecursive(outputDir);
     }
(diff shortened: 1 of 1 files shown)


RikkiGibson:
The thing to note here is that more specific config settings (e.g. `csharp.clear-output-folder` vs `clear-output-folder`) have higher precedence, and settings provided on the command line break a tie with settings provided by README.md. I think this raises the question of whether to simply prepend `csharp.` to all our config settings to protect from similar changes down the line.

RikkiGibson:
@hovsepm Please have a look and merge if you think this is a reasonable change.

RikkiGibson:
Need to tweak to point to new azure-rest-api-specs branch.

src/ResourceManagement/gulpfile.js
```diff
@@ -6,7 +6,7 @@ var spawn = require('child_process').spawn;
 var fs = require('fs');
 
 const mappings = require('./api-specs.json');
-const defaultSpecRoot = "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current";
+const defaultSpecRoot = "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master";
```
hovsepm: why do we move from current to master?

RikkiGibson: azure-rest-api-specs people decided to delete the "current" branch today and put all the recent stuff into master.

hovsepm: @RikkiGibson then please also do open the similar PR in the azure-sdk-for-java repo. Thanks!

RikkiGibson:
Original PR for Java which fixed the output-folder problem:
https://github.com/Azure/azure-libraries-for-java/pull/162

Second PR for Java which fixes the current/master branch problem:
https://github.com/Azure/azure-libraries-for-java/pull/163

Issue #104, opened 2017-11-27
Unauthorised response when calling Graph API methods using local PFX file authentication


alanmk:
I'm currently unable to use Graph API functionality when authenticating via local PFX files. 

It looks like the problem's [here](https://github.com/Azure/azure-libraries-for-net/blob/367c89ff76a31016fa2ea5f278022fd3d9b0dc83/src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs#L128) i.e. the `ProcessHttpRequestAsync `method of `AzureCredentials `is not passing the `ActiveDirectoryServiceSettings `object into `LoginSilentAsync`.

hovsepm:
fix should be available in v1.7

PR #166, opened 2018-02-07
Fixed issue where netstandard doesn't get the correct token for the graph endpoint

graemefoster:
Without this the token sent to the graph endpoint doesn't work on .net standard.
Looks like an oversight in the code. See issue here:

https://github.com/Azure/azure-libraries-for-net/issues/104


hovsepm:
@azuresdkci test this please.

hovsepm:
@graemefoster Thanks for the fix!

graemefoster:
@hovsepm you’re welcome - do you know when this will be released on nuget? 

Thanks,
Graeme

hovsepm:
@graemefoster our initial plan is to make 1.7 available sometime around beginning of March.

PR #186, opened 2018-02-27
Removed dependency of 7z copy during the build.

hovsepm:


hovsepm:
Sign job: http://azuresdkci.cloudapp.net/view/NetCore-SDK/job/Fluent-NetCore-SDK-Sign/19

Issue #170, opened 2018-02-08
BatchAIJob.Refresh() does not work


lenala:

hovsepm:
Fixed in v1.7

PR #187, opened 2018-02-27
Fixing BatchAIJob refresh functionality

lenala:
fix for https://github.com/Azure/azure-libraries-for-net/issues/170


hovsepm:
@lenala could you add a test to make sure Refresh works?

PR #188, opened 2018-02-28
Removing client side constraint on service-bus queue and topic name max size

anuchandy:
…


psytraxx:
The same happens on SubscriptionsOperations - in the same namespace - is there a change we can remove this limit there as well?

PR #195, opened 2018-03-04
Support custom\external ServiceClientCredentials

eric-winkler:
A number of the managers downcast the `RestClient`'s `Credentials` to `AzureCredentials` (eg https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/Compute/ComputeManager.cs#L46).  Combined with the public facing side of the API accepting `ServiceClientCredentials` as input, we end up with an unfortunate violation of Liskov Substitution Principal that means code like this doesn't work;

```
// using Microsoft.Azure.Services.AppAuthentication(Prerelease)
var azureServiceTokenProvider = new AzureServiceTokenProvider();
var token = await azureServiceTokenProvider.GetAccessTokenAsync("https://management.core.windows.net/");
var tenantId = azureServiceTokenProvider.PrincipalUsed.TenantId;

var client = RestClient
	.Configure()
	.WithEnvironment(AzureEnvironment.AzureGlobalCloud)
	.WithLogLevel(HttpLoggingDelegatingHandler.Level.Basic)
	.WithCredentials(new TokenCredentials(token))
	.Build();

// throws "Unable to cast object of type 'Microsoft.Rest.TokenCredentials' to type 'Microsoft.Azure.Management.ResourceManager.Fluent.Authentication.AzureCredentials'."
var azure = Azure
	.Authenticate(client, tenantId)
	.WithDefaultSubscription();
```

This change moves us "all in" on the `AzureCredentials` usage on the `RestClient` and provides consumers a new (working) way to inject custom `ServiceClientCredentials` via a new constructor on `AzureCredentials`.

This is a breaking change from an API signature perspective, but as per the example above, I expect any code that was supplying ServiceClientCredentials would be broken anyway.

### DIFF
--- a/src/ResourceManagement/Compute/ComputeManager.cs
+++ b/src/ResourceManagement/Compute/ComputeManager.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.Management.Compute.Fluent
         {
             storageManager = StorageManager.Authenticate(restClient, subscriptionId);
             networkManager = NetworkManager.Authenticate(restClient, subscriptionId);
-            rbacManager = GraphRbacManager.Authenticate(restClient, ((AzureCredentials)(restClient.Credentials)).TenantId);
+            rbacManager = GraphRbacManager.Authenticate(restClient, restClient.Credentials.TenantId);
         }
 
         #endregion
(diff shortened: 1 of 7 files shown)


hovsepm:
@azuresdkci test this please

hovsepm:
Thanks for the contribution. This will be available from release v1.9

PR #208, opened 2018-03-08
Uploading an app service's certificate straight from a byte[]

eric-winkler:
This change introduces a `WithPfxByteArrayToUpload` to allow for directly attaching certificates from their byte[] content. eg;
```
var certBytes = certificateService.RetreiveCertificate();
webapp
	.Update()
	.DefineSslBinding()
	.ForHostname("my.hostname")
	.WithPfxByteArrayToUpload(certBytes, "password")
	.WithSniBasedSsl()
	.Attach()
	.Apply();
```

**Background:**
There didn't seem to be a mechanism to upload an SSL certificate for a webapp without first persisting it to disk (ie https://github.com/Azure/azure-libraries-for-net/blob/0e725531319775e348c21540c9a2c25000106c32/src/ResourceManagement/AppService/HostNameSslBindingImpl.cs#L110).  In situations where certificate management is API driven (eg Let's Encrypt) proxying via disk is an inconvenience.


hovsepm:
@azuresdkci test this please

src/ResourceManagement/AppService/HostNameSslBindingImpl.cs
```diff
@@ -66,6 +66,18 @@ private string GetCertificateThumbprint(string pfxPath, string password)
             throw new KeyNotFoundException("Private key not found in the pfx file");
         }
 
+        private static string GetCertificateThumbprint(byte[] pfxByteArray, string password)
+        {
+            var collection = new X509Certificate2Collection();
```
martincostello: Why not just use `new X509Certificate2(rawData, password)` to get the thumbprint, rather than loading it into a collection first?

eric-winkler: I've followed the existing implementation in `GetCertificateThumbprint(string pfxPath, string password)` as I was unsure if there was some nuance to the particular approach previously taken.

Happy to update my method and the existing one if there is consensus that the `X509Certificate2Collection` is unnecessary.

martincostello: Ah OK. I was just curious, as I came across this PR while trying to see if there was a existing issue for a problem I'm having related to certificates, and the constructor approach is what I do in my particular app.

eric-winkler: On closer inspection I don't think the existing implementation is actually behaving as intended - enumerating a X509Certificate2Collection import doesn't provide any guarantees about the presence of a private key

Thanks for highlighting this martin, I'll put something together to address.

eric-winkler:
Have rebased against latest master and addressed some feedback that ultimately uncovered a preexisting bug in which a missing private key inside a pfx container would not trigger the expected exception.

hovsepm:
@azuresdkci retest this please.

The-First-Tiger:
Hey, ist this actually rolling out?

yaohaizh:
@weidongxu-microsoft please take a second look for this PR when it is good to merge now. If everything is okay, go ahead to merge the PR. Thanks.

weidongxu-microsoft:
We probably want to use X509KeyStorageFlags.EphemeralKeySet, however it is not available on 4.6.

Checking whether we should use X509KeyStorageFlags.MachineKeySet. DefaultKeySet might not work for temporay profile. Probably for later consideration.

eric-winkler:
I'm keen to hear what you discover @weidongxu-microsoft - the previous implementation used `PersistKeySet` which just seems wrong for the transient nature of this cert.  I have not tested this on an azure webapp.

weidongxu-microsoft:
@eric-winkler I think it is good enough. I will merge it after a makeshift automation get pass with the new interface.

weidongxu-microsoft:
@eric-winkler Merged to 1.27.1, release would be Oct. 25th. Thanks for your contribution!

PR #227, opened 2018-03-23
CosmosDB fixed null primaryReadonlyMasterKey and secondaryReadonlyMas…

olandese:
CosmosDB, fixed null primaryReadonlyMasterKey and secondaryReadonlyMasterKey.

The problem is probably coming from the documentation, where the client code is generated from:

https://docs.microsoft.com/en-us/rest/api/documentdbresourceprovider/databaseaccounts/listkeys

the API 2015-04-08 is returning the following object back: 
{
  "primaryMasterKey": "theprimarykey",
  "secondaryMasterKey": "thesecondarykey",
  "primaryReadonlyMasterKey": "theprimaryReadonlyMasterKey",
  "secondaryReadonlyMasterKey": "thesecondaryReadonlyMasterKey"
}

and NOT as described:
{
  "primaryMasterKey": "primaryMasterKey",
  "secondaryMasterKey": "secondaryMasterKey",
  "properties": {
    "primaryReadonlyMasterKey": "primaryReadonlyMasterKey",
    "secondaryReadonlyMasterKey": "secondaryReadonlyMasterKey"
  }
}

### DIFF
--- a/src/ResourceManagement/CosmosDB/Generated/Models/DatabaseAccountListKeysResultInner.cs
+++ b/src/ResourceManagement/CosmosDB/Generated/Models/DatabaseAccountListKeysResultInner.cs
@@ -73,13 +73,13 @@ namespace Microsoft.Azure.Management.CosmosDB.Fluent.Models
         /// <summary>
         /// Gets base 64 encoded value of the primary read-only key.
         /// </summary>
-        [JsonProperty(PropertyName = "properties.primaryReadonlyMasterKey")]
+        [JsonProperty(PropertyName = "primaryReadonlyMasterKey")]
         public string PrimaryReadonlyMasterKey { get; private set; }
 
         /// <summary>
         /// Gets base 64 encoded value of the secondary read-only key.
         /// </summary>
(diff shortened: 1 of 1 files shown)


hovsepm:
@azuresdkci test this please.

milismsft:
@olandese we might in a be better state if we instead refresh the "inner" classes using the latest AutoREST swagger.

olandese:
if the swagger from the REST API is now OK, no problem. But as I described in the PR the documentation for the api-version=2015-04-08 is wrong, and if this documentation is also coming from a wrong swagger the problem will not be solved. 
If you refresh against the right swagger or a newer REST API that will be ok. 
It would be nice to get the fix in the next 2 - 3 weeks, in the meanwhile I will call the REST API from C# directly.

milismsft:
Thank you for the contribution to this repo. We hope that https://github.com/Azure/azure-libraries-for-net/pull/247 which was recently merged already addresses the changes in this PR. If that is not the case please open an issue and we will try to fix/address it ASAP.

PR #232, opened 2018-03-23
SQL: add new Fluent features

milismsft:
Added Failover groups, DNS aliases, SQL MSI and server secret, virtual networks, capabilities, general metrics and usage metrics, check names and others


src/ResourceManagement/Sql/SqlChildrenOperationsImpl.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            this.DeleteBySqlServer(ResourceUtils.GroupFromResourceId(id),
+                ResourceUtils.NameFromResourceId(ResourceUtils.ParentRelativePathFromResourceId(id)),
+                ResourceUtils.NameFromResourceId(id));
+
+        }
+
+        ///GENMHASH:4D33A73A344E127F784620E76B686786:5C3E0E7E1CF165D67A5D6B44C259B469
+        public async Task DeleteByIdAsync(string id, CancellationToken cancellationToken = default(CancellationToken))
+        {
```
hovsepm: Starting from this line and further down the code - please pass cancellationToken to the method async calls.

src/ResourceManagement/Sql/SqlDatabaseAutomaticTuningImpl.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        ///GENMHASH:93F5525F475C77754B229151C3005F4B:FE981C7CC77129BC45DD1FFCFCD2402F
+        public async Task<Microsoft.Azure.Management.Sql.Fluent.ISqlDatabaseAutomaticTuning> ApplyAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            this.Inner.Options = this.automaticTuningOptionsMap.Count == 0 ? null : this.automaticTuningOptionsMap;
+            var databaseAutomaticTuningInner = await this.sqlServerManager.Inner.DatabaseAutomaticTuning
+                .UpdateAsync(this.resourceGroupName, this.sqlServerName, this.sqlDatabaseName, this.Inner, cancellationToken);
+            this.SetInner(databaseAutomaticTuningInner);
+            this.automaticTuningOptionsMap = (databaseAutomaticTuningInner != null && databaseAutomaticTuningInner.Options != null) ? new Dictionary<string, AutomaticTuningOptions>(databaseAutomaticTuningInner.Options) : new Dictionary<string, AutomaticTuningOptions>();
+            return databaseAutomaticTuningInner != null ? this : null;
```
hovsepm: I'm not sure I understood what are you trying to do here. Why would you return null in any case within ApplyAsync? Throwing exception on error should be preferred way, right?

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
```diff
@@ -159,6 +157,22 @@ public SqlDatabaseImpl WithEdition(string edition)
             return this;
         }
 
+        ///GENMHASH:376C162D9E950FA8A14B198790B45E34:5BF104F4489A5E42BAC27A2492EFC800
+        public IReadOnlyList<Microsoft.Azure.Management.Sql.Fluent.ISqlDatabaseUsageMetric> ListUsageMetrics()
```
hovsepm: This method seems to be not implemented. Please check.

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
```diff
@@ -205,6 +219,16 @@ public SqlDatabaseImpl WithActiveDirectoryLoginAndPassword(string administratorL
             return this;
         }
 
+        ///GENMHASH:B793713C1E6696E957FDFDD58A692C32:6F6FD3F1EE6142188DD9CD51E89E8125
+        public ISqlDatabaseAutomaticTuning GetDatabaseAutomaticTuning()
```
hovsepm: this as well - not implemented.

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
```diff
@@ -626,6 +650,27 @@ public SqlDatabaseImportRequestImpl ImportBacpac(IStorageAccount storageAccount,
                 .ImportFrom(storageAccount, containerName, fileName);
         }
 
+        ///GENMHASH:D07614EC964723EC7E563278F8FD0BE6:9903ECC9F8F5837D63A44DAA11F79C8E
+        public async Task<Microsoft.Azure.Management.Sql.Fluent.ISqlDatabaseUsageMetric> ListUsageMetricsAsync(CancellationToken cancellationToken = default(CancellationToken))
```
hovsepm: ok, so it looks like the entire file is not implemented :)

src/ResourceManagement/Sql/SqlDatabaseOperationsImpl.cs
```diff
@@ -182,6 +182,41 @@ public ISqlDatabase Get(string name)
...
+            {
+                var inner = await this.sqlServerManager.Inner.Databases
+                    .GetAsync(sqlServer.ResourceGroupName, sqlServer.Name, name, cancellationToken: cancellationToken);
+                return inner != null ? new SqlDatabaseImpl(inner.Name, (SqlServerImpl)sqlServer, inner, sqlServer.Manager) : null;
+            }
+            catch (CloudException ex) when (ex.Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+            catch (AggregateException ex) when ((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound)
```
hovsepm: >.InnerExceptions[0] [](start = 51, length = 19)

this is kind of unsafe - what if AggregatedException will not have any inner exceptions?

src/ResourceManagement/Sql/SqlElasticPoolOperationsImpl.cs
```diff
@@ -183,6 +199,25 @@ public ISqlElasticPool Get(string name)
...
+            {
+                var inner = await sqlServer.Manager.Inner.ElasticPools
+                    .GetAsync(sqlServer.ResourceGroupName, sqlServer.Name, name, cancellationToken);
+                return inner != null ? new SqlElasticPoolImpl(inner.Name, (SqlServerImpl)sqlServer, inner, sqlServer.Manager) : null;
+            }
+            catch (CloudException ex) when (ex.Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+            catch (AggregateException ex) when ((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound)
```
hovsepm: >((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound) [](start = 47, length = 101)

same comment. your condition is unsafe. Please revise.

src/ResourceManagement/Sql/SqlFailoverGroupImpl.cs
```diff
@@ -0,0 +1,430 @@
...
+            if (innerObject != null && innerObject.Id != null)
+            {
+                try
+                {
+                    ResourceId resourceId = ResourceId.FromString(innerObject.Id);
+                    this.resourceGroupName = resourceId.ResourceGroupName;
+                    this.sqlServerName = resourceId.Parent.Name;
+                    this.sqlServerLocation = innerObject.Location;
+                }
+                catch (Exception) { }
```
hovsepm: what is the point of catching all in this simple case?

src/ResourceManagement/Sql/SqlFailoverGroupImpl.cs
```diff
@@ -0,0 +1,430 @@
...
+        {
+            //$ SqlFailoverGroupImpl self = this;
+            if (this.Inner.Location == null)
+            {
+                this.Inner.Location = this.sqlServerLocation;
+            }
+            var failoverGroupInner = await this.sqlServerManager.Inner.FailoverGroups
+                .CreateOrUpdateAsync(this.resourceGroupName, this.sqlServerName, this.Name(), this.Inner, cancellationToken);
+            this.SetInner(failoverGroupInner);
+            return failoverGroupInner != null ? this : null;
```
hovsepm: >return failoverGroupInner != null ? this : null; [](start = 12, length = 48)

same thing - always return "this".

src/ResourceManagement/Sql/SqlFirewallRuleOperationsImpl.cs
```diff
@@ -174,6 +189,25 @@ public ISqlFirewallRule Get(string name)
...
+                var firewallRuleInner = await sqlServer.Manager.Inner.FirewallRules
+                    .GetAsync(sqlServer.ResourceGroupName, sqlServer.Name, name, cancellationToken);
+                return firewallRuleInner != null ? new SqlFirewallRuleImpl(firewallRuleInner.Name, (SqlServerImpl)sqlServer, firewallRuleInner, sqlServer.Manager) : null;
+            }
+            catch (CloudException ex) when (ex.Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+            catch (AggregateException ex) when ((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
```
hovsepm: unsafe condition. Please revise.

src/ResourceManagement/Sql/SqlServerDnsAliasImpl.cs
```diff
@@ -0,0 +1,214 @@
...
+            this.name = name;
+            this.sqlServerManager = sqlServerManager;
+            if (innerObject != null && innerObject.Id != null)
+            {
+                try
+                {
+                    ResourceId resourceId = ResourceId.FromString(innerObject.Id);
+                    this.resourceGroupName = resourceId.ResourceGroupName;
+                    this.sqlServerName = resourceId.Parent.Name;
+                } catch (Exception) { }
```
hovsepm: > } catch (Exception) { } [](start = 15, length = 24)

please remove catch all.

src/ResourceManagement/Sql/SqlServerKeyImpl.cs
```diff
@@ -0,0 +1,303 @@
...
+            this.sqlServerManager = sqlServerManager;
+            if (innerObject != null && innerObject.Id != null)
+            {
+                try
+                {
+                    ResourceId resourceId = ResourceId.FromString(innerObject.Id);
+                    this.resourceGroupName = resourceId.ResourceGroupName;
+                    this.sqlServerName = resourceId.Parent.Name;
+                }
+                catch (Exception) { }
```
hovsepm: >atch (Exception) { } [](start = 17, length = 20)

no catch all please.

src/ResourceManagement/Sql/SqlServerKeyImpl.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        ///GENMHASH:C608B2A3A894C122878626FA09D94406:BE33F46EDC996DB65B7F46EC68568C24
+        public SqlServerKeyImpl WithAzureKeyVaultKey(string uri)
+        {
+            this.Inner.ServerKeyType = Models.ServerKeyType.AzureKeyVault.Value;
+            this.Inner.Uri = uri;
+            // If the key URI is "https://YourVaultName.Vault.Azure.Net/keys/YourKeyName/01234567890123456789012345678901",
+            // then the Server Key Name should be formatted as: "YourVaultName_YourKeyName_01234567890123456789012345678901"
+            string[] items = uri.Split('/');
+            this.serverKeyName = $"{items[2].Split('.')[0]}_{items[4]}_{items[5]}";
```
hovsepm: >items[2].Split('.')[0] [](start = 36, length = 22)

did you check this? AFAIK you cannot have  complex calculation in this kind of string formatting. May need to assign this to a variable and then use.

src/ResourceManagement/Sql/SqlServerKeyImpl.cs
```diff
@@ -0,0 +1,303 @@
...
+            return Microsoft.Azure.Management.ResourceManager.Fluent.Core.Region.Create(this.Inner.Location);
+        }
+
+        ///GENMHASH:0202A00A1DCF248D2647DBDBEF2CA865:9C622B7760E8AA499BCF909CBA2DEBA0
+        public async Task<Microsoft.Azure.Management.Sql.Fluent.ISqlServerKey> CreateResourceAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var serverKeyInner = await this.sqlServerManager.Inner.ServerKeys
+                .CreateOrUpdateAsync(this.resourceGroupName, this.sqlServerName, this.serverKeyName, this.Inner, cancellationToken);
+            this.SetInner(serverKeyInner);
+            return serverKeyInner != null ? this : null;
```
hovsepm: >return serverKeyInner != null ? this : null; [](start = 12, length = 44)

always return "this"

src/ResourceManagement/Sql/SqlVirtualNetworkRuleOperationsImpl.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                var vnetInner = await this.sqlServerManager.Inner.VirtualNetworkRules
+                    .GetAsync(resourceGroupName, sqlServerName, name, cancellationToken);
+                return vnetInner != null ? new SqlVirtualNetworkRuleImpl(resourceGroupName, sqlServerName, vnetInner.Name, vnetInner, sqlServerManager) : null;
+            }
+            catch (CloudException ex) when (ex.Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+            catch (AggregateException ex) when ((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound)
```
hovsepm: >gregateException ex) when ((ex.InnerExceptions[0] as CloudExcepti [](start = 21, length = 65)

unsafe - please revise.

src/ResourceManagement/Sql/SqlVirtualNetworkRuleOperationsImpl.cs
```diff
@@ -0,0 +1,201 @@
...
+                var vnetInner = await sqlServer.Manager.Inner.VirtualNetworkRules
+                    .GetAsync(sqlServer.ResourceGroupName, sqlServer.Name, name, cancellationToken);
+                return vnetInner != null ? new SqlVirtualNetworkRuleImpl(vnetInner.Name, (SqlServerImpl)sqlServer, vnetInner, sqlServerManager) : null;
+            }
+            catch (CloudException ex) when (ex.Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+            catch (AggregateException ex) when ((ex.InnerExceptions[0] as CloudException).Response.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
```
hovsepm: same here - please revise

src/ResourceManagement/Sql/SqlVirtualNetworkRuleOperationsImpl.cs
```diff
@@ -0,0 +1,201 @@
...
+            return (this.sqlServer != null) ? result.WithExistingSqlServer(this.sqlServer) : result;
+        }
+
+        public ISqlVirtualNetworkRule GetById(string id)
+        {
+            return Extensions.Synchronize(() => this.GetByIdAsync(id));
+        }
+
+        public Task<ISqlVirtualNetworkRule> GetByIdAsync(string id, CancellationToken cancellationToken = default(CancellationToken))
+        {
```
hovsepm: you have bunch of NotImplemented methods here. Please revise.

src/ResourceManagement/Sql/SqlDatabaseOperationsImpl.cs
milismsft: I opened https://github.com/Azure/azure-libraries-for-net/issues/233 since this needs to be also fixed in the "core" classes.

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
milismsft: Fixed

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
milismsft: Fixed

src/ResourceManagement/Sql/SqlDatabaseAutomaticTuningImpl.cs
milismsft: Fixed

src/ResourceManagement/Sql/SqlChildrenOperationsImpl.cs
milismsft: Fixed

src/ResourceManagement/Sql/SqlFailoverGroupImpl.cs
milismsft: If the inner freshly created, it does not have a valid ID.

src/ResourceManagement/Sql/SqlServerDnsAliasImpl.cs
milismsft: same as the previous case, inner.Id might not be available if the inner is constructed rather than retrieved from the service

src/ResourceManagement/Sql/SqlServerDnsAliasOperationsImpl.cs
```diff
@@ -0,0 +1,169 @@
...
+            ResourceId resourceId = ResourceId.FromString(oldSqlServerId);
+            await this.sqlServerManager.Inner.ServerDnsAliases
+                .AcquireAsync(resourceId.ResourceGroupName, resourceId.Name, dnsAliasName, newSqlServerId + DNS_ALIASES + dnsAliasName, cancellationToken);
+        }
+
+        ///GENMHASH:6296EBA2A3987F54D7F72FBAD6479C42:43BF008A6256BDEA0A296AB514FC7683
+        public async Task AcquireAsync(string dnsAliasName, string sqlServerId, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            await this.AcquireAsync(this.sqlServer.ResourceGroupName, this.sqlServer.Name, dnsAliasName, sqlServerId);
+        }
```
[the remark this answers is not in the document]
milismsft: fixed

src/ResourceManagement/Sql/SqlServerKeyImpl.cs
milismsft: yes, I checked ;-)

src/ResourceManagement/Sql/SqlServerKeyImpl.cs
milismsft: fixed

src/ResourceManagement/Sql/SqlVirtualNetworkRuleOperationsImpl.cs
milismsft: fixed

src/ResourceManagement/Sql/SqlDatabaseImpl.cs
milismsft: fixed

src/ResourceManagement/Sql/CheckNameAvailabilityResultImpl.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///GENMHASH:6E11439BC0EE92936D1373B518E18596:0FD5BCAE10307382B777A749B232A151
+        internal CheckNameAvailabilityResultImpl(CheckNameAvailabilityResponseInner inner) : base(inner)
+        {
+        }
+
+        ///GENMHASH:ECE9AA3B22E6D72ED037B235766E650D:853A4A58250437949396E2FC3A58AA01
+        public bool IsAvailable()
+        {
+            return this.Inner.Available.GetValueOrDefault(false);
+        }
```
hovsepm: I do not think this pattern is good. If the inner type is nullable then our wrapped type also should be nullable. It is not good to return default value - in this case false, when server did not have this field in the response body. IMHO there is a difference between null and false value. And this logic is present all over SQL. If this is a new pattern I'd ask you to change it in this release.

src/ResourceManagement/Sql/SqlDatabaseUsageMetricImpl.cs
```diff
@@ -0,0 +1,65 @@
...
+        ///GENMHASH:3E38805ED0E7BA3CAEE31311D032A21C:61C1065B307679F3800C701AE0D87070
+        public string Name()
+        {
+            return this.Inner.Name;
+        }
+
+        ///GENMHASH:9D196E486CC1E35756FD0BEDAB3F3BE4:1A97C3177E4C3DBBD400096F05CC2999
+        public double Limit()
+        {
+            return this.Inner.Limit.GetValueOrDefault(0.0d);
```
hovsepm: >            return this.Inner.Limit.GetValueOrDefault(0.0d) [](start = 0, length = 59)

my comment about having nullible in the wrapper is better shown here - so if I'd not set value for Limit during create you will return me 0.0 after create?

src/ResourceManagement/Sql/SqlServerAutomaticTuningImpl.cs
```diff
@@ -0,0 +1,159 @@
...
+            return Inner.DesiredState.GetValueOrDefault();
+        }
+
+        ///GENMHASH:93F5525F475C77754B229151C3005F4B:F84ED1C7DC13C78E0925116541403988
+        public async Task<Microsoft.Azure.Management.Sql.Fluent.ISqlServerAutomaticTuning> ApplyAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var serverAutomaticTuningInner = await this.sqlServerManager.Inner.ServerAutomaticTuning
+                .UpdateAsync(this.resourceGroupName, this.sqlServerName, this.Inner, cancellationToken);
+            this.SetInner(serverAutomaticTuningInner);
+            return serverAutomaticTuningInner != null ? this : null;
```
hovsepm: >return serverAutomaticTuningInner != null ? this : null; [](start = 12, length = 56)

This method should return "this".

src/ResourceManagement/Sql/CheckNameAvailabilityResultImpl.cs
milismsft: The return must be a boolean in this case, and "null" should be treated as "false". This API checks if the server name is available (SQL names are FQDN across the whole cloud).

src/ResourceManagement/Sql/SqlDatabaseUsageMetricImpl.cs
milismsft: You cannot set a "limit" here; it's a read only property which should be consistent with what we return in Java.

src/ResourceManagement/Sql/SqlServerAutomaticTuningImpl.cs
milismsft: Hmm... I thought I got them all... Fixed now.

Tests/Fluent.Tests/Sql/SqlTests.cs
```diff
@@ -72,6 +73,323 @@ public void CanOperateSqlFromRollUpClient()
...
+                Assert.Equal(AutomaticTuningOptionModeDesired.Default, serverAutomaticTuning.TuningOptions["forceLastGoodPlan"].DesiredState);
+
+                var databaseAutomaticTuning = dbFromSample.GetDatabaseAutomaticTuning();
+                Assert.Equal(4, databaseAutomaticTuning.TuningOptions.Count);
+
+                // The following results in "InternalServerError" at the moment
+                databaseAutomaticTuning.Update()
+                    .WithAutomaticTuningMode(AutomaticTuningMode.Auto)
+                    .WithAutomaticTuningOptions("createIndex", AutomaticTuningOptionModeDesired.Off)
+                    .WithAutomaticTuningOptions("dropIndex", AutomaticTuningOptionModeDesired.On)
```
hovsepm: design thought: you are setting one option here right? shouldn't method name be in a single form - WithAutomaticTuningOption?

src/ResourceManagement/Sql/CheckNameAvailabilityResultImpl.cs
milismsft: FYI we actually could do worst; check out Traffic Manager  CheckProfileDnsNameAvailabilityResult()

hovsepm:
@milismsft I'm done with review. Please address comments.

Tests/Fluent.Tests/Sql/SqlTests.cs
milismsft: The option name comes from a Dictionary which does not contain a static set of keys; it's a service returned map.

milismsft: sorry... missed the plural vs singular. Yes, it can be singular

src/ResourceManagement/Sql/CheckNameAvailabilityResultImpl.cs
hovsepm: @milismsft  Seems like another issue for github :) Could you add it? I'm done with the PR review. If you addressed all the comments and know what to do - you can merge :).

milismsft: Returning "false" for nullable bool or "0" for nullable integer, long or double is not new. We need to stay consistent with the Java implementation; whatever we return in Java we should return as default here as well. For instance if the return in Java type is "Integer" which can be "null", we return 0, not "null".

milismsft: I should add that for string, UUID/Guid or DateTime which are not primary types, we should always return the nullable type (in Java we return null too). In SQL I do this for the newer stuff but for the old stuff I could not change them because of risk of introducing a breaking change.

Issue #209, opened 2018-03-08
Creating a SQL reource fails on macOS


wis3guy:
In a netcore1.1  console app, this snippet:

```c#
azure.SqlServers.Define(n)
   .WithRegion(region)
   .WithExistingResourceGroup(rg)
   .WithAdministratorLogin(sqlLogin)
   .WithAdministratorPassword(sqlPassword)
   .Create()
```
... works as advertised when run on Windows (f.ex. from Visual Studio), but on macOS the following exception is thrown:

```
Microsoft.Rest.SerializationException: Unable to deserialize the response. ---> Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: <. Path '', line 0, position 0.
   at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonTextReader.Read()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)
   at Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject[T](String json, JsonSerializerSettings settings)
   at Microsoft.Azure.Management.Sql.Fluent.FirewallRulesOperations.<CreateOrUpdateWithHttpMessagesAsync>d__5.MoveNext()
   --- End of inner exception stack trace ---
   at Microsoft.Azure.Management.Sql.Fluent.FirewallRulesOperations.<CreateOrUpdateWithHttpMessagesAsync>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[...]

PR #248, opened 2018-04-05
SQL fix for Firewall Rule

milismsft:
Work around for SQL Firewall Rule service returning XML response rather than expected JSON.


hovsepm:
fix for https://github.com/Azure/azure-libraries-for-net/issues/209

PR #259, opened 2018-04-11
Changes for new .NET generator

jianghaolu:


src/ResourceManagement/ResourceManager/Core/ProxyResource.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// Resource Id
+        /// </summary>
+        [JsonProperty(PropertyName = "id")]
+        public virtual string Id { get; protected set; }
+
+        /// <summary>
+        /// Resource name
+        /// </summary>
+        [JsonProperty(PropertyName = "name")]
+        public string Name { get; private set; }
```
hovsepm: why do we have Id as protected set but name and type as private set?

PR #260, opened 2018-04-11
Update RBAC to 2018-01-01-preview

jianghaolu:


hovsepm:
you have the same ExpandableStringEnumConverter class in the pr https://github.com/Azure/azure-libraries-for-net/pull/259 , right? should this pr be rebased after that one will be  merged?

PR #267, opened 2018-04-12
Updating Network to 2018-02

lenala:


hovsepm:
Tests\Fluent.Tests\Network\NetworkWatcherTests.cs, line 84
```
                ITopology topology = nw.Topology()
                    .WithTargetResourceGroup(vm0.ResourceGroupName)
                    .Execute();
```
I think Topology should be property type and not a method.

hovsepm:
same thing in Samples\Network\ManageNetworkWatcher.cs, line 187 - Topology seems to be property instead of method.

hovsepm:
Please revert ProxyResource changes - setter on Id should be protected.

PR #270, opened 2018-04-13
Upgrading CRT and CRT.Autnetication to the latest available version.

hovsepm:


hovsepm:
http://azuresdkci.cloudapp.net/view/NetCore-SDK/job/fluent-samples-nigthly/347/

PR #277, opened 2018-04-18
Regenerate network with native subresource

jianghaolu:


src/ResourceManagement/Network/Generated/Models/ApplicationGatewayInner.cs
```diff
@@ -241,12 +239,6 @@ public ApplicationGatewayInner(string location = default(string), string id = de
         [JsonProperty(PropertyName = "etag")]
         public string Etag { get; set; }
 
-        /// <summary>
-        /// Gets or sets the list of availability zones denoting where the resource needs to come from.
-        /// </summary>
-        [JsonProperty(PropertyName = "zones")]
-        public IList<string> Zones { get; set; }
-
```
hovsepm: This seems to be wrong

PR #308, opened 2018-05-03
Fixed System.Runtime wrong dependency in .Net 4.5.2

hovsepm:


hovsepm:
sign-job: http://azuresdkci.cloudapp.net/view/NetCore-SDK/job/Fluent-NetCore-SDK-Sign/27/

PR #337, opened 2018-05-23
private build for CosmosDB customer validation

milismsft:


milismsft:
Signed job http://azuresdkci.cloudapp.net/view/NetCore-SDK/job/Fluent-NetCore-SDK-Sign/29/

PR #357, opened 2018-06-12
Copy session records to fileshare when running live tests

lenala:


lenala:
Already merged as a part of another PR.

PR #368, opened 2018-06-22
JWT payload not padded in FunctionCredentials

shep1987:
We ran into an issue where the payload of the JWT couldn't be parsed correctly, this turned out to be because it needed to be padded, this fix should work, but may be better to look into the use of a jwt parsing library.


hovsepm:
@azuresdkci test this please.

hovsepm:
@jianghaolu could you review this PR?

hovsepm:
Thanks for the contribution. This fix will be available in release 1.13.

Issue #356, opened 2018-06-11
Get service principal fails when service principal has multiple keys with the same name


LockTar:
We tried this with v1.10 and updated it to 1.11.1 but still with the samen result.

When we get the details of a service principal, we get an error `An item with the same key has already been added`.
We use the code: `azure.ActiveDirectoryApplications.GetByNameAsync(name);`

We found the issue around this. Our service principals has multiple keys with the same name.
If we remove one key it succeeds. So the problem is with **multiple keys** with the **same name**.

hovsepm:
@LockTar fix has been merged. Will be shipped as part of release 1.13 due around mid-July.

hovsepm:
Release 1.13 is out. Please upgrade your references.

PR #373, opened 2018-06-27
Fixing operations with ActiveDirectoryApplication with multiple credentials with same name

lenala:
Fixing issue #356  Get service principal fails when service principal has multiple keys with the same name

Note: this is a breaking change as getters for credentials previously returned Dictionary indexed by name and now return Dictionary indexed by id.


hovsepm:
@lenala this is a breaking change that needs to be properly documented in the release notes.

PR #381, opened 2018-07-03
Virtual machine: support for Azure hybrid benefit

Enjeej:
Added support for Azure hybrid benefit for Virtual Machines by adding the possibility to set the license-type during create or update of the VM definition.


hovsepm:
@azuresdkci test this please

src/ResourceManagement/Compute/Domain/InterfaceImpl/VirtualMachineImpl.cs
```diff
@@ -645,6 +645,16 @@ VirtualMachine.Update.IUpdate VirtualMachine.Update.IUpdate.WithSize(VirtualMach
...
         }
 
+        /// <summary>
+        /// Specifies the license-type for the virtual machine.
+        /// </summary>
+        /// <param name="licenseType">The license-type that should be used for the virtual machine.</param>
+        /// <return>The next stage of the definition.</return>
+        VirtualMachine.Update.IUpdate VirtualMachine.Update.IUpdate.WithLicenseType(string licenseType)
+        {
+            return this.WithLicenseType(licenseType) as VirtualMachine.Update.IUpdate;
```
lenala: According to swagger spec description, this value cannot be updated: https://github.com/Azure/azure-rest-api-specs/blob/c8516b913b3c420f38e036c3e2b8dbf1550cb089/specification/compute/resource-manager/Microsoft.Compute/stable/2018-04-01/compute.json#L5796

Enjeej: I pretty sure the update also worked when I tested this, but since it is already a couple of weeks ago I am not 100% sure about it anymore. On the other hand in the Azure Portal you can enable and disable "Azure Hybrid Benefit" as many times as you want.  

So I tested it again. I am able to set and empty the license-type multiple times for the same VM. However, I did not check if I could set another license-type after I used: "Windows_Server".
But I guess when this is not supported the server will return an exception.

Enjeej:
@lenala I've removed the casts

Enjeej:
Hi @lenala ,

I get a the following build error:

c:\workspace\Fluent-SdkCI\build.proj : error MSB4057: The target "Clean" does not exist in the project.

Build FAILED.
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.04

any idea what is going wrong?

Best regards,

NJ

hovsepm:
@azuresdkci retest this please

lenala:
Hi @Enjeej, our build process has changed, please see https://github.com/Azure/azure-libraries-for-net#prerequisites to build locally - it's one time setup.

Enjeej:
Hi @lenala, locally it build without problems. It was here on GitHub where the problem lay. 
However, it think it has been resolved, since you approved the PR.

Thank you for reviewing and approving my pull requests. 

@hovsepm, thank you for the support

PR #396, opened 2018-07-12
Fix for "unable to set Augmented Rules on Network Security Groups"

lenala:


hovsepm:
@please add test coverage

PR #406, opened 2018-07-18
Update to Compute 2018-06-01

lenala:


lenala:
Replaced with https://github.com/Azure/azure-libraries-for-net/pull/412

PR #469, opened 2018-09-20
fixing testFx null ref exception during recording

shahabhijeet:
Fix for null ref during recording of test.
Adding new version of HttpRecorder and TestFx


hovsepm:
Addressed in PR https://github.com/Azure/azure-libraries-for-net/pull/469

PR #548, opened 2018-12-10
Add API to add an IpSecurityRestriction while Defining a WebApp

dpetillo:
I needed this functionality added to the API.  If it can be added to master, I think ill need advisement on my API signature(s) and how to properly regenerate the model:

src\ResourceManagement\AppService\Generated\Models\IpSecurityRestriction.cs

The copy in this commit was taken from https://github.com/Azure/azure-sdk-for-net and lightly modified.


hovsepm:
@azuresdkci test this please.

src/ResourceManagement/AppService/Generated/Models/IpSecurityRestriction.cs
```diff
@@ -29,13 +21,30 @@ public IpSecurityRestriction()
         /// Initializes a new instance of the IpSecurityRestriction class.
         /// </summary>
         /// <param name="ipAddress">IP address the security restriction is
-        /// valid for.</param>
+        /// valid for.
```
praries880: @dpetillo 
you can undo these changes, The model got updated with these properties added to it :
https://github.com/Azure/azure-libraries-for-net/blob/303240ff67225bf622b29cdaa0a6268e37bb5c63/src/ResourceManagement/AppService/Generated/Models/IpSecurityRestriction.cs#L51

praries880:
@dpetillo 
Thanks for the PR :)

We have already updated the models to include the changes you have here. We still need to add a public API to make IPSecurityRestriction available on the Fluent API though. We will get try to get the suggestions here in webapps in the near future. 

The API changes you made do look good, but we still have to look into the update (PATCH) scenario and add tests for the same.

yaohaizh:
@ChenTanyi please follow with this PR. Thanks.

PR #550, opened 2018-12-10
Fix for CosmosDB argument null exception.

hovsepm:


hovsepm:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/view/Fluent%20SDK/job/Fluent-Net-SDK-Sign/16/

PR #598, opened 2019-02-19
Fix indentified char in sample script

jianghaolu:


jianghaolu:
license check stuck again. Re-open.

PR #638, opened 2019-04-05
Fix for : 617 Add user assigned msi to web app

praries880:
Fix for : #617


jianghaolu:
> Skip tests failing despite files being present 

It's probably because you did not set the properties for the pfx files to be "copy if newer"?

praries880:
> > Skip tests failing despite files being present
> 
> It's probably because you did not set the properties for the pfx files to be "copy if newer"? like here: https://github.com/Azure/azure-libraries-for-net/blob/master/Samples/Samples.csproj#L326


@jianghaolu Took care of the review comment. Thanks for suggesting the fix.

praries880:
@jianghaolu / @lmazuel 
I am merging these changes in since the only comment was WRT the tests which i have since fixed.

PR #668, opened 2019-05-20
Switch to using a common version property

weshaggard:
cc @praries880


Directory.Build.props
```diff
@@ -1,24 +1,40 @@
...
+    <GenerateAssemblyProductAttribute>true</GenerateAssemblyProductAttribute>
+    <GenerateAssemblyCopyrightAttribute>true</GenerateAssemblyCopyrightAttribute>
+    <GenerateNeutralResourcesLanguageAttribute>true</GenerateNeutralResourcesLanguageAttribute>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <FluentVersion>1.22.0</FluentVersion>
+    <VersionPrefix>$(FluentVersion)</VersionPrefix>
+
+    <!-- All the AssemblyInfo files had the Assembly Version set to 1.0.0.60 unless there is a strong reason for that I wouldn't explicitly set it differently -->
```
weshaggard: @praries880 do you know if there is a reason the assembly version was fixed at 1.0.0.60? If so we can uncomment this line to make them match. If not then we can just delete these line completely.

praries880: @weshaggard I am looking into it. Personally i do not think we should stick to the one version(as per your suggestion). I am looking into why have we done this and if there is some specific reason to sticking to the one version.

praries880: @weshaggard 
Can you update the comment to indicate that we rev the assembly version only when there is a breaking change in the SDK... otherwise we retain the same version as before.

This  is what we have been doing so far n i dont wanna change it as of now.

PR #669, opened 2019-05-20
Add ADO build config files.

praries880:
This will enable us to use ADO for CI and release.


.azure-pipelines/mgmt.yml
```diff
@@ -0,0 +1,37 @@
+pr:
```
weshaggard: I would rename the mgmt.yml to represent what you are building. Perhaps fluent?

.azure-pipelines/mgmt.yml
```diff
@@ -0,0 +1,37 @@
...
+
+variables:
+  DotNetCoreVersion: '2.1.503'
+
+jobs:
+  - job: Windows_Build
+    pool:
+      vmImage: 'VS2017-Win2016'
+    variables:
+      IgnoreBuildTokens: ''
```
weshaggard: If you don't use this you can just remove it here and in the msbuildARguments below.

weshaggard:
Can you send me a link to a test run you have kicked off?

.azure-pipelines/fluentNetSdk.yml
```diff
@@ -0,0 +1,37 @@
+pr:
+  branches:
+    include:
+    - master
+    - search-preview
```
weshaggard: These branches don't look interesting to your repo. I would remove all but master.

praries880: done

PR #694, opened 2019-06-07
ACI: v2018_10 Update

iscai-msft:


src/ResourceManagement/ContainerInstance/ContainerGroupImpl.cs
```diff
@@ -106,9 +179,9 @@ public string ProvisioningState()
             return this.Inner.ProvisioningState;
         }
 
-        public OSTypeName OSType()
+        public OperatingSystemTypes OSType()
         {
-            return Fluent.OSTypeName.Parse(this.Inner.OsType);
+            return OperatingSystemTypes.Parse(this.Inner.OsType);
```
anuchandy: This is a breaking change i.e. change in return type. We should revert the deleted ` Fluent.OSTypeName.cs` file then we can just revert this method.

PR #731, opened 2019-07-16
 Update compute to version 2019-03-01

yaohaizh:

PR #743, opened 2019-07-22
Update compute to version 2019-03-01

yaohaizh:


yaohaizh:
Re merge the PR since to clean the commits history: 
https://github.com/Azure/azure-libraries-for-net/pull/731

PR #760, opened 2019-08-09
support list function envelopes

weidongxu-microsoft:


src/ResourceManagement/AppService/Domain/IFunctionApp.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Azure.Management.AppService.Fluent
     using Microsoft.Azure.Management.Storage.Fluent;
     using System.Collections.Generic;
     using System.IO;
+    using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
```
yaohaizh: Sort using

src/ResourceManagement/AppService/Domain/InterfaceImpl/FunctionAppImpl.cs
```diff
@@ -22,6 +22,7 @@ namespace Microsoft.Azure.Management.AppService.Fluent
     using Microsoft.Azure.Management.Storage.Fluent;
     using System.Collections.Generic;
     using System;
+    using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
```
yaohaizh: Same here

src/ResourceManagement/AppService/Domain/InterfaceImpl/FunctionEnvelopeImpl.cs
```diff
@@ -0,0 +1,153 @@
...
+    internal partial class FunctionEnvelopeImpl
+    {
+        string IHasId.Id
+        {
+            get
+            {
+                return this.Id();
+            }
+        }
+
```
yaohaizh: Make sure the comments consistent behavior.

src/ResourceManagement/AppService/FunctionEnvelopeImpl.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        /// <summary>
+        /// Gets the name of the resource
+        /// </summary>
+        public string Name()
+        {
+            return this.Inner.Name;
+        }
+
```
yaohaizh: Either add all the comments or not.

weidongxu-microsoft:
Sure. I will be more careful on comments (or better to write some script to generate these files next time).

BTW, does Impl.cs need comments in it?

src/ResourceManagement/AppService/Domain/IFunctionApp.cs
```diff
@@ -95,5 +96,21 @@ public interface IFunctionApp  :
         /// <param name="functionName">The name of the function.</param>
         /// <return>The function key.</return>
         Task<System.Collections.Generic.IReadOnlyDictionary<string,string>> ListFunctionKeysAsync(string functionName, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// List the functions
+        /// </summary>
+        IEnumerable<IFunctionEnvelope> ListFunctions();
```
yaohaizh: Should be IReadonlyList?

weidongxu-microsoft: IEnumerable is already readonly for the underlying collection.

I am referring to this interface on the method signature of returned resource list (my feeling is that it more like an operation on object, than a property of object):

`    public interface ISupportsListing<T>
    {
        /// <summary>
        /// Lists all the resources of the specified type in the currently selected subscription.
        /// </summary>
        /// <returns>list of resources</returns>
        IEnumerable<T> List();

        /// <summary>
        /// Lists all the resources of the specified type in the currently selected subscription.
        /// </summary>
        /// <returns>list of resources</returns>
        Task<IPagedCollection<T>> ListAsync(bool loadAllPages = true, CancellationToken cancellationToken = default(CancellationToken));
    }
`

I will change to IReadOnlyList if you feel it be more consistent with the codebase.

yaohaizh: For API, it is important to provide consistent interface/API exposed to the customer.

Issue #485, opened 2018-10-05
Fix Functions on App Service Plan logic


m-mertz:
It feels like the Functions logic in its current form was written for the consumption plan model, and app service plan support was tacked on without proper testing. There are a few issues pointing at specific symptoms of this, but no "this needs an overhaul" type issue, so here goes.

The scenario I’d like to see enabled:

```csharp
IFunctionApp app = await appServiceManager.FunctionApps
    .Define(appName)
    .WithExistingAppServicePlan(appServicePlan)
    .WithExistingResourceGroup(resourceGroup)
    .WithExistingStorageAccount(storageAccount)
    .WithRuntimeVersion("~2")
    .CreateAsync()
    .ConfigureAwait(false);
```

Expected outcome: A new function app with the same app settings as a new app service plan function app created in the portal, and no weird side effects.

Gaps:

1. SDK creates WEBSITE_CONTENTAZUREFILECONNECTIONSTRING app setting.
2. SDK creates WEBSITE_CONTENTSHARE app setting.
3. SDK creates file share in the specified storage account (the file share referenced in setting above). I don’t just want the setting to disappear, I’d also like to avoid creating garbage data in the storage account.
4. When retrieving the function app again in the SDK, it should recognize the configured storage account by AzureWebJobsStorage setting alone (if running on an app service plan).

The mentioned settings are for consumption plans only: https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#websitecontentazurefileconnectionstring

Related issues: #457, #265

xenalite:
@hovsepm @jianghaolu I just started using fluent APIs for Function Apps and hit this exact issue and the ones linked in related. One of the original issues has been open for over a year. Are there any plans to resolve these?
I attempted to get around this problem by provisioning the function app using a generic ResourceManager, but I cannot specify `Kind` property to be `functionApp`.

yaohaizh:
Fixed by PR#761.

Issue #457, opened 2018-09-04
Adding app settings to existing Function App creates a new storage account


david-jarman:
Creating a new storage account is an unintended consequence of calling

```CSharp
functionApp.Update()
.WithAppSetting("key", "value")
.Apply();
```
This also causes the currently deployed functions to be lost since the function app is now pointing at a new storage account.

The offending code is in [FunctionAppImpl.cs](https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/AppService/FunctionAppImpl.cs) Line 163-166

I think the root cause of this issue is that the currentStorageAccount is never populated with the actual storage account that is associated with the function app

david-jarman:
My hope was that the function app methods in the fluent SDK would be idempotent. As I've been looking through the code, it seems that this might not always be the case.

I thought I could work around this issue by adding the original connect storage account to the Update() using WithExistingStorageAccount, but that method ends up updating the WEBSITE_CONTENTSHARE app setting to a random resource name on Apply(), Now that I've debugged the issue, I've ended up with a bunch of different file shares in that storage account, and my functions that were deployed get wiped out.

I've been using fluent for a while, but functions are the first service I've used that offers methods that aren't idempotent.

m-mertz:
Having the same issue. Just wanted to update an app setting and then the SDK tries to create a new storage account. Even fails in my case because the generated storage account name is too long (my function app name is 20 chars long).

In my case, the function app is on an app service plan. According to [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings), WEBSITE_CONTENTAZUREFILECONNECTIONSTRING and WEBSITE_CONTENTSHARE are only required for consumption plan apps. But it looks like the SDK needs these settings to recognize that an app already has a storage account.

Strongly related to this, I don't think these settings should be defined at all when creating a new function app on an app service plan, but they are. Quick repro (inspect app settings of new app):

```csharp
IFunctionApp newFunctionApp = await azure.AppServices.FunctionApps
    .Define("functionAppName")
    .WithExistingAppServicePlan(appServicePlan)
    .WithExistingResourceGroup("resourceGroup")
    .WithExistingStorageAccount(storageAccount)
    .CreateAsync()
    .ConfigureAwait(false);
```

m-mertz:
There's a simple workaround for app settings, and possibly other properties that are inherited from web apps. This works because a Function App is just a special Web App, so we can use those methods instead:

[...]

PR #761, opened 2019-08-12
fix for function app, app setting for storage account

weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-net/issues/485
If app service plan is not consumption, skip WEBSITE_CONTENTAZUREFILECONNECTIONSTRING and WEBSITE_CONTENTSHARE

https://github.com/Azure/azure-libraries-for-net/issues/457
If it is Update (NOT Create) operation, and storage account is not set, do not create new storage account (and do not update related app setting).


src/ResourceManagement/AppService/FunctionAppImpl.cs
```diff
@@ -154,17 +160,31 @@ internal override async Task<Models.SiteInner> SubmitAppSettingsAsync(SiteInner
             }
         }
 
+        private bool IsConsumptionAppServicePlan(PricingTier pricingTier)
+        {
+            SkuDescription description = pricingTier.SkuDescription;
```
yaohaizh: Null check

src/ResourceManagement/AppService/FunctionAppImpl.cs
```diff
@@ -137,13 +137,19 @@ internal override async Task<Models.SiteInner> SubmitAppSettingsAsync(SiteInner
...
+                bool isConsumptionPlan = IsConsumptionAppServicePlan(servicePlan.PricingTier);
+
                 var keys = await storageAccountToSet.GetKeysAsync(cancellationToken);
                 var connectionString = string.Format("DefaultEndpointsProtocol=https;AccountName={0};AccountKey={1}",
                     storageAccountToSet.Name, keys[0].Value);
                 WithAppSetting("AzureWebJobsStorage", connectionString);
                 WithAppSetting("AzureWebJobsDashboard", connectionString);
-                WithAppSetting("WEBSITE_CONTENTAZUREFILECONNECTIONSTRING", connectionString);
-                WithAppSetting("WEBSITE_CONTENTSHARE", SdkContext.RandomResourceName(Name, 32));
+                if (isConsumptionPlan)
```
yaohaizh: Doesn't need additional variable here.

src/ResourceManagement/AppService/FunctionAppImpl.cs
```diff
@@ -154,17 +160,31 @@ internal override async Task<Models.SiteInner> SubmitAppSettingsAsync(SiteInner
...
+                {
+                    WithNewConsumptionPlan();
+                }
+                if (currentStorageAccount == null && storageAccountToSet == null && storageAccountCreatable == null)
+                {
+                    WithNewStorageAccount(SdkContext.RandomResourceName(Name, 20).Replace("-", String.Empty), Storage.Fluent.Models.SkuName.StandardGRS);
+                }
+            } else
             {
-                WithNewStorageAccount(SdkContext.RandomResourceName(Name, 20).Replace("-", String.Empty), Storage.Fluent.Models.SkuName.StandardGRS);
```
yaohaizh: The else is not necessary. Please remove it to keep the code clean.

src/ResourceManagement/AppService/FunctionAppImpl.cs
```diff
@@ -154,17 +160,31 @@ internal override async Task<Models.SiteInner> SubmitAppSettingsAsync(SiteInner
...
+                // for Update+Apply, skip check null and create new app service plan and storage account, use existing ones if caller does not specify otherwise
             }
-            return await base.CreateAsync(cancellationToken);
+            return await base.CreateAsync(cancellationToken, multiThreaded);
```
yaohaizh: What's the reason to add the multiThreaded here?

weidongxu-microsoft: method signature is
`public override async Task<IFunctionApp> CreateAsync(CancellationToken cancellationToken = default(CancellationToken), bool multiThreaded = true)`

so I guess previous base call miss that multiThreaded param.

PR #764, opened 2019-08-19
Network: update to 2019-06

yaohaizh:


src/ResourceManagement/Network/ApplicationGatewayImpl.cs
```diff
@@ -1008,6 +1012,51 @@ public ApplicationGatewayImpl WithInstanceCount(int capacity)
...
+            return this;
+        }
+
+        public ApplicationGatewayImpl WithWebApplicationFirewall(ApplicationGatewayWebApplicationFirewallConfiguration config)
+        {
+            this.Inner.WebApplicationFirewallConfiguration = config;
+            return this;
+        }
+
+        public ApplicationGatewayImpl withAutoScale(int minCapacity, int maxCapacity)
```
weidongxu-microsoft: seems same method as line 1018?

yaohaizh: Removed this methods.

PR #766, opened 2019-08-21
fix cosmosdb

weidongxu-microsoft:
In short, the root cause is that manager.Inner.BaseUri = https://management.azure.com/ (which should not even be there) is hiding the real one (manager.Inner as FluentServiceClientBase<CosmosDB>).BaseUri = https://management.usgovcloudapi.net/

### DIFF
--- a/Tests/Fluent.Tests/ManagementClientTest.cs
+++ b/Tests/Fluent.Tests/ManagementClientTest.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Xunit;
+using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
+using Microsoft.Azure.Management.ResourceManager.Fluent;
+using Microsoft.Azure.Management.ResourceManager.Fluent.Authentication;
+using Microsoft.Azure.Management.Fluent;
+using Microsoft.Rest.Azure;
+using System;
+
+namespace Fluent.Tests
+{
+    public class ManagementClientTest
+    {
+        [Fact]
+        public void ManagementClientBaseUri()
+        {
+            AzureEnvironment environment = AzureEnvironment.AzureUSGovernment;
+            AzureCredentials credentials = new AzureCredentialsFactory().FromServicePrincipal("mockId", "mockSecret", "mockTenant", environment);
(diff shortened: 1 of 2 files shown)


weidongxu-microsoft:
Client

> Would you be able to find out if there has been any recent regression in using the Azurefluent libraries for Cosmos Db against the Azure Gov environment. An API I have not changed to query the Comsos DB during my deployment has stopped working on Gov environments but still succeeds on Commercial environments. In particular on object of type Microsoft.Azure.Management.CosmosDB.Fluent.ICosmosDBAccounts call to method GetByResourceGroupAsync is failing and returning null. I have confirmed in Azure portal the Resource Group Name and Cosmos DB name I am sending to function call matches exactly what I see in the DB. The Service Principal this is executing in has Contributor access to the Cosmos DB it is failing to query. Thank you.
> 
> Actually decided to Use Fiddler and step through some of the code and I think I have an idea why this is happening. And needs to be a bug on Microsoft fluent API for Cosmos.
> I am passing AzureUSGovernment as the AzureEnvironment when I log in. However for IAzure.CosmosDBAccounts it is putting in a base URI of https://management.azure.com/when for all other objects on the IAzure interface the BaseUri is set to https://management.usgovcloudapi.net/ (edited) 
> This explains why works for commercial but not gov. Seems there is a regression in Microsoft.Azure.Management.CosmosDB.Fluent that the CosmosDBAccounts is not getting the correct BaseUri set for AzureGov environment.
> 
> I first noticed this with latest (1.24.1)

Reply

> Confirm that this is a regression on C# Fluent SDK (probably since end of June).
> 
> Observation from Citrix is correct.
> 
> 
> We will be releasing version 1.25 with the fix in a couple of days.
> 
> 
> If Citrix need an immediate fix, here is a workaround:
> 
> After initialization of ICosmosDBManager instance (either through Azure.Authenticate or through CosmosDBManager.Authenticate), add this line:
> 
> manager.Inner.BaseUri = (manager.Inner as FluentServiceClientBase<CosmosDB>).BaseUri;
> 
> manager is of type ICosmosDBManager (can be obtained via azure.CosmosDBAccounts.Manager)
> 
> In short, the root cause is that manager.Inner.BaseUri = https://management.azure.com/ (which should not even be there) is hiding the real one (manager.Inner as FluentServiceClientBase<CosmosDB>).BaseUri = https://management.usgovcloudapi.net/

PR #771, opened 2019-08-25
Update new versions of the stacks

acelina:
Adds new versions of NodeJS, NETCORE, PHP, and Ruby which are currently available in Azure config lists

### DIFF
--- a/src/ResourceManagement/AppService/RuntimeStack.cs
+++ b/src/ResourceManagement/AppService/RuntimeStack.cs
@@ -7,6 +7,21 @@ namespace Microsoft.Azure.Management.AppService.Fluent
     /// </summary>
     public partial class RuntimeStack
     {
+        public static readonly RuntimeStack NodeJS_LTS = new RuntimeStack("NODE", "LTS");
+        public static readonly RuntimeStack NodeJS_10_LTS = new RuntimeStack("NODE", "10 LTS");
+        public static readonly RuntimeStack NodeJS_10_14 = new RuntimeStack("NODE", "10.14");
+        public static readonly RuntimeStack NodeJS_10_12 = new RuntimeStack("NODE", "10.12");
+        public static readonly RuntimeStack NodeJS_10_10 = new RuntimeStack("NODE", "10.10");
+        public static readonly RuntimeStack NodeJS_10_1 = new RuntimeStack("NODE", "10.1");
+        public static readonly RuntimeStack NodeJS_9_4 = new RuntimeStack("NODE", "9.4");
+        public static readonly RuntimeStack NodeJS_8_12 = new RuntimeStack("NODE", "8.12");
+        public static readonly RuntimeStack NodeJS_8_11 = new RuntimeStack("NODE", "8.11");
+        public static readonly RuntimeStack NodeJS_8_9 = new RuntimeStack("NODE", "8.9");
+        public static readonly RuntimeStack NodeJS_8_8 = new RuntimeStack("NODE", "8.8");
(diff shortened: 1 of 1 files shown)


yaohaizh:
@acelina Thanks for your contribution.  @weidongxu-microsoft could you help take a look at this file to see whether we need update app service spec. Thanks.

src/ResourceManagement/AppService/RuntimeStack.cs
```diff
@@ -7,6 +7,21 @@ namespace Microsoft.Azure.Management.AppService.Fluent
     /// </summary>
     public partial class RuntimeStack
     {
+        public static readonly RuntimeStack NodeJS_LTS = new RuntimeStack("NODE", "LTS");
+        public static readonly RuntimeStack NodeJS_10_LTS = new RuntimeStack("NODE", "10 LTS");
```
weidongxu-microsoft: I am not sure whether Azure support this version "10 LTS" (it is not on portal). Would you refer me to the source of the config lists?

weidongxu-microsoft: Seems should be "10-lts"? (result from "az webapp list-runtimes")

acelina: Thanks for the good catch. I had a wrong reference list. 

Updated the list using ```az webapp list-runtimes --linux```. But now the question is, what about the Windows stack. It gives a different result? Is this class OS dependent?

weidongxu-microsoft: I think on current code design, withBuiltInImage(RuntimeStack) works for Linux only (it continues on flow after withExistingLinuxPlan or withNewLinuxPlan). Windows would initial a different flow, which continues as withPythonVersion etc.

Please give me some more time to confirm it and try out some test cases myself.

Thanks.

src/ResourceManagement/AppService/RuntimeStack.cs
```diff
@@ -15,15 +32,26 @@ public partial class RuntimeStack
...
         public static readonly RuntimeStack PHP_7_0 = new RuntimeStack("PHP", "7.0");
+        public static readonly RuntimeStack PHP_7_2 = new RuntimeStack("PHP", "7.2");
+        public static readonly RuntimeStack PHP_7_3 = new RuntimeStack("PHP", "7.3");
         public static readonly RuntimeStack NETCore_V1_0 = new RuntimeStack("DOTNETCORE", "1.0");
         public static readonly RuntimeStack NETCore_V1_1 = new RuntimeStack("DOTNETCORE", "1.1");
-        public static readonly RuntimeStack Ruby_2_3 = new RuntimeStack("RUBY", "2.3");
+        public static readonly RuntimeStack NETCore_V2_0 = new RuntimeStack("DOTNETCORE", "2.0");
+        public static readonly RuntimeStack NETCore_V2_1 = new RuntimeStack("DOTNETCORE", "2.1");
+        public static readonly RuntimeStack NETCore_V2_2 = new RuntimeStack("DOTNETCORE", "2.2");
+        public static readonly RuntimeStack Ruby_2_3 = new RuntimeStack("RUBY", "2.3.8");
```
weidongxu-microsoft: We would use major version here. So for RUBY we will have 2.3, 2.4, 2.5, 2.6.
The reason is that minor version might change too quick for us to update in code.

With major version, server would figure out the mirror version.

acelina: Fixed. Thanks!

weidongxu-microsoft:
@acelina Thanks for your contribution.

Will be available on 1.26

Issue #466, opened 2018-09-18
Deleting last app in app service plan deletes plan, no way to override


m-mertz:
Deleting the last app in an app service plan also deletes the plan. This behavior is consistent throughout portal and other surfaces. But in portal or cli, for example, it can be overridden, see e.g. https://github.com/Azure/azure-cli/issues/4497 (--keep-empty-plan).

In Fluent SDK, there doesn't seem to be a way to override this so that the app service plan stays around. Relevant methods (not complete list):

```csharp
appServiceManager.WebApps.DeleteByIdAsync
appServiceManager.WebApps.DeleteByResourceGroupAsync
appServiceManager.FunctionApps.DeleteByIdAsync
appServiceManager.FunctionApps.DeleteByResourceGroupAsync
```

ppittle:
Any updates on this?

weidongxu-microsoft:
Fixed in 1.27. To be released on Oct. 11th.

Interface:
https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/AppService/Domain/IWebApps.cs

Sample code
`appServiceManager.WebApps.DeleteById(webApp2.Id, null, false);`

PR #781, opened 2019-09-11
additional option when deleting webapp

weidongxu-microsoft:


yaohaizh:
Is there any issue/features related to this PR?

weidongxu-microsoft:
@yaohaizh
https://github.com/azure/azure-libraries-for-net/issues/466

src/ResourceManagement/AppService/WebAppDeleteOption.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Management.AppService.Fluent
+{
+    using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
+
+    /// <summary>
+    /// Defines options for WebApp deletion.
+    /// </summary>
+    public class WebAppDeleteOption : IBeta
```
yaohaizh: Should be app service delete option

yaohaizh: Second though, should we remove this class to keep the interface consistent and simple to use. 

Provide default on the interface level.

src/ResourceManagement/AppService/ISupportWebAppDeleteOption.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Microsoft.Azure.Management.AppService.Fluent
+{
+    using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Provides access to deleting a resource from Azure, with additional option.
+    /// </summary>
+    public interface ISupportWebAppDeleteOption : IBeta
```
yaohaizh: ISupportAppServiceDelete

src/ResourceManagement/AppService/WebAppDeleteOption.cs
weidongxu-microsoft: OK. Interface updated to use bool? .

Issue #784, opened 2019-09-12
[FEATURE REQ] Update AKS to use the latest swagger spec


yaohaizh:
Support feature:  
- [ ] VMSS support: https://github.com/Azure/AKS/issues/945#issuecomment-523220680

PR #793, opened 2019-09-18
Container Service: update to 2019 08 01

ChenTanyi:
#784


weidongxu-microsoft:
any test case on newly added agent pool?

ChenTanyi:
> any test case on newly added agent pool?

@weidongxu-microsoft yeah, in the last commit, create kubernetes cluster with agent pool type set.

PR #805, opened 2019-09-23
Batch: Update swagger to the latest

xccc-msft:


Tests/Fluent.Tests/Batch/BatchTests.cs
```diff
@@ -11,6 +11,9 @@
 using Xunit;
```
yaohaizh: Sort using

src/ResourceManagement/Batch/ApplicationPackagesImpl.cs
```diff
@@ -59,7 +59,7 @@ protected override IList<ApplicationPackageImpl> ListChildResources()
                     childResources.Add(new ApplicationPackageImpl(applicationPackage.Name, parent, applicationPackage, client));
                 }
             } catch (CloudException) {
-
+                //ignore the exception when creating application package during application definition stage
```
yaohaizh: What is the reason to ignore the exception?

src/ResourceManagement/Batch/Domain/BatchAccount/Definition/IDefinition.cs
```diff
@@ -2,19 +2,13 @@
...
-    using Microsoft.Azure.Management.Batch.Fluent;
-    using Microsoft.Azure.Management.ResourceManager.Fluent.Core.Resource.Definition;
-    using Microsoft.Azure.Management.Batch.Fluent.Application.Definition;
-    using Microsoft.Azure.Management.ResourceManager.Fluent.Core.GroupableResource.Definition;
-
     /// <summary>
     /// The stage of a Batch account definition allowing the adding of a Batch application or creating the Batch account.
     /// </summary>
     public interface IWithCreateAndApplication :
-        Microsoft.Azure.Management.Batch.Fluent.BatchAccount.Definition.IWithCreate,
```
yaohaizh: What is the reason to remove the full qualified name?

src/ResourceManagement/Batch/ApplicationPackagesImpl.cs
xccc-msft: The exception will come when it calls batchAccount.Update().DefineNewApplication(appId).DefineNewApplicationPackage(appPackageName).Attach().Apply(). The request to list all application packages will get response of "the specific application does not exist" as we will commit and sync up both the application and application package at same time. What we have done on Java side is that we register error for CloudException so it return empty child resource list in the same situation. Thus, we need to do similar thing in .net side to catch the exception and return empty list.

src/ResourceManagement/Batch/Domain/BatchAccount/Definition/IDefinition.cs
xccc-msft: It's unnecessary to keep full qualified name here. This namespace is from IDefinition.cs itself.

Tests/Fluent.Tests/Batch/BatchTests.cs
```diff
@@ -242,5 +245,135 @@ public async Task CanCreateBatchAccountWithApplication()
...
         }
+
+        [Fact]
+        public async Task CanCreateBatchAccountWithPool()
+        {
+            using (var context = FluentMockContext.Start(this.GetType().FullName))
+            {
+                try
+                {
+                    var poolId = "testPool";
```
yaohaizh: Please use random name for testing

src/ResourceManagement/Batch/Domain/IBatchAccount.cs
```diff
@@ -61,6 +61,16 @@ public interface IBatchAccount :
         /// </summary>
         ProvisioningState ProvisioningState { get; }
 
+        /// <summary>
+        ///Gets a value indicating whether the core quota for the Batch Account is enforced per Virtual Machine family or not.
```
yaohaizh: Space after `///`

Tests/Fluent.Tests/Batch/BatchTests.cs
xccc-msft: Should we apply the random name rule to existing test cases in BatchTests.cs? It seems all names there are fixed value rather than generating randomly. We don't have random generator in .net while we do have that in Java for only resource group name, batch name, storage account name.

xccc-msft: For now CanCRUDBatchAccounts() generated the name randomly while CanCreateBatchAccountWithApplication() didn't. I will change to random name for it and the new added test case.

yaohaizh: We have random string generator for other test cases. Please check other test cases.

Tests/Fluent.Tests/Batch/BatchTests.cs
xccc-msft: done

Tests/Fluent.Tests/Batch/BatchTests.cs
xccc-msft: done

src/ResourceManagement/Batch/Domain/IBatchAccount.cs
xccc-msft: done

PR #825, opened 2019-10-09
Compute: support hypervisor generation for disk

xccc-msft:


src/ResourceManagement/Compute/Domain/IDisk.cs
```diff
@@ -32,6 +32,11 @@ public interface IDisk :
         /// </summary>
         Models.OperatingSystemTypes? OSType { get; }
 
+        /// <summary>
+        /// Gets the hypervisor generation of the Virtual Machine.
```
yaohaizh: Should be disk?

src/ResourceManagement/Compute/Domain/InterfaceImpl/DiskImpl.cs
```diff
@@ -306,6 +317,16 @@ Disk.Update.IUpdate Disk.Update.IWithOSSettings.WithOSType(OperatingSystemTypes
             return this.WithOSType(osType);
         }
 
+        /// <summary>
```
yaohaizh: Same as the above

src/ResourceManagement/Compute/Domain/IDisk.cs
xccc-msft: The description is provided by service team within DiskInner. The hypervisor generation is related to virtual machine while it is applicable to OS disk only.

weidongxu-microsoft:
Should have same in definition flow?

PR #827, opened 2019-10-09
Compute: support reimage for virtual machine

xccc-msft:


weidongxu-microsoft:
Add one line test if possible.

PR #829, opened 2019-10-10
WithoutReplyUrl must add existing Reply URLs

Johan-Lindvall-Stratsys:
WithoutReplyUrl must add existing Reply URLs before attempting to remove one (see WithReplyUrl in update mode for similar logic).

If not added, the updateParameters.ReplyUrls will be null and nothing will happen.

### DIFF
--- a/src/ResourceManagement/Graph.RBAC/ActiveDirectoryApplicationImpl.cs
+++ b/src/ResourceManagement/Graph.RBAC/ActiveDirectoryApplicationImpl.cs
@@ -281,10 +281,11 @@ namespace Microsoft.Azure.Management.Graph.RBAC.Fluent
 
         public ActiveDirectoryApplicationImpl WithoutReplyUrl(string replyUrl)
         {
-            if (updateParameters.ReplyUrls != null)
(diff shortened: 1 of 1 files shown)


src/ResourceManagement/Graph.RBAC/ActiveDirectoryApplicationImpl.cs
```diff
@@ -281,10 +281,11 @@ public Uri SignOnUrl()
 
         public ActiveDirectoryApplicationImpl WithoutReplyUrl(string replyUrl)
         {
-            if (updateParameters.ReplyUrls != null)
+            if (updateParameters.ReplyUrls == null)
             {
-                updateParameters.ReplyUrls.Remove(replyUrl);
+                updateParameters.ReplyUrls = new List<string>(this.Inner.ReplyUrls);
```
ChenTanyi: I think since ReplyUrls is an ISet type, HashSet is better than List

ChenTanyi:
@Johan-Lindvall-Stratsys thanks for contribution.

Issue #796, opened 2019-09-18
Not all regions listed in Microsoft.Azure.Management.ResourceManager.Fluent.Core.Region


MiroslavVanicky:
Hi,

when looking into Microsoft.Azure.Management.ResourceManager.Fluent.Core.Region there are public fields for regions (eg, public static readonly Region USWest2 = new Region("westus2")).

Why are there some regions missing, eg Switzerland regions? (List is not beeing synchronized with current Azure regions?)

Thank You

weidongxu-microsoft:
Fixed in 1.27.1, to be released on Oct. 25th.

PR #843, opened 2019-10-18
Add missed Azure region values

weidongxu-microsoft:


yaohaizh:
@weidongxu-microsoft please associate with the issue if possible when create PR for better tracking and provide meaningful PR title. Thus, we can get track the history and commit message. Thanks.

qub1n:
This issue 
https://github.com/Azure/azure-libraries-for-net/issues/796

weidongxu-microsoft:
image

eh.. so I am the only one who can see this?

Tests/Fluent.Tests/ResourceManager/SubscriptionsTests.cs
```diff
@@ -65,5 +68,31 @@ public async Task CanGetByIdAsync()
                 Assert.Equal(subscription.SubscriptionId.ToLowerInvariant(), azure.SubscriptionId.ToLowerInvariant());
             }
         }
+
+        [Fact(Skip = "Util to generate missing regions")]
+        public void GenerateMissingRegion()
```
yaohaizh: Should we skip this test case?

weidongxu-microsoft: It is actually util code to faciliate generating code when azure should add new region in the future.

It is no use in replay mode, it had to run in live. Hence the SKIP.

Issue #840, opened 2019-10-16
Clearer support for how to authenticate with User Assigned Managed Identity


apilosofms:
## Problem
Today I spent ages trying to get auth to work with a User Assigned Managed Identity. The problem was that I was not specifying the identity ID when creating the `MSILoginInformation`. That's fine when using a Managed Service Identity, but apparently not with a User Assigned one.

However, there's no way to know that; the API doesn't guide you in the right direction because a constructor which takes an ID doesn't exist.

## Recommended solution
Consider methods like this in `MSILoginInformation` (possibly renamed to `ManagedIdentityInformation`):
- `MSILoginInformation.FromManagedServiceIdentity()`
- `MSILoginInformation.FromUserAssignedIdentityByClientId(string clientId)`
- `MSILoginInformation.FromUserAssignedIdentityByObjectId(string objectId)`
- `MSILoginInformation.FromUserAssignedIdentityByResourceId(string resourceId)`

Along with those, remove the default ctor if possible.
It might make more sense to move these methods one level up, to something like `AzureCredentialsFactory`.

## Alternatives
You could make ctor overloads to hint to the user that there are other bits of information that might be required - but there are two complications:
- The overloads would use the same inputs (single string), requiring some hoop jumping.
- Not as clear to the user which to use.

Either way, as long as it's obvious to the user that using a User Assigned Managed Identity requires extra info (unlike a Managed Service Identity), that's all that's required. Have them fall into the pit of success...

yaohaizh:
@apilosofms thanks for reporting this!

PR #849, opened 2019-10-21
feat: add more options to create MSI Login Credentials

ChenTanyi:
#840


weidongxu-microsoft:
My suggestion is that it have 4 or less (some might be merged if same set of parameters) factory method like:

FromAppServiceSystemAssigned
FromAppServiceUserAssigned
FromVirtualMachineSystemAssigned
FromVirtualMachineUserAssigned

So from method name one would know which to use.

Pending for discussion. @yaohaizh

ChenTanyi:
I don't know whether there is difference between AppService and VirtualMachine to login in. But I think if we separate resource type, it would get larger and larger when other resource need to access azure services.

weidongxu-microsoft:
Of course it will get bigger. That is the point. Else I only need one curl, why library.

Issue #852, opened 2019-10-22
[BUG] ListFunctionKeys - authentication failure


xenalite:
**Describe the bug**
401 Unauthorized response code when calling ListFunctionKeys

***Exception or Stack Trace***
```
Microsoft.Rest.Azure.CloudException: Operation returned an invalid status code 'Unauthorized'
   at Microsoft.Azure.Management.AppService.Fluent.FunctionAppImpl.ListFunctionKeysAsync(String functionName, CancellationToken cancellationToken)
   at Microsoft.Azure.Management.ResourceManager.Fluent.Core.Extensions.Synchronize[TResult](Func`1 function)
   at Microsoft.Azure.Management.AppService.Fluent.FunctionAppImpl.ListFunctionKeys(String functionName)
   at WTW.AML.DeploymentTool.Actions.PrepareTestStageActions.Execute() in C:\workspace\AML\build\AML.DeploymentTool\WTW.AML.DeploymentTool\Actions\PrepareTestStageActions.cs:line 75
```

***Fiddler traces***
Request 1:
```
GET https://management.azure.com/subscriptions/SubscriptionId/resourceGroups/AML-DEV-CORE/providers/Microsoft.Web/sites/ObfuscatedFunctionName/functions/admin/token?api-version=2018-02-01 HTTP/1.1
x-ms-client-request-id: d3e64f94-ea78-43b0-ac02-b9a907aed89a
Accept-Language: en-US
Authorization: Bearer ey...
User-Agent: FxVersion/4.6.28008.02 OSName/Windows OSVersion/Microsoft.Windows.10.0.16299. Microsoft.Azure.Management.AppService.Fluent.WebSiteManagementClient/1.0.0.60 MacAddressHash/64a540467fa6a85fb8bf50f9bf6bc5f917ffc453337600b5c271b6524602b07b
Host: management.azure.com
```

Response 1:
```
HTTP/1.1 200 OK
Cache-Control: no-cache
Pragma: no-cache
Content-Length: 317
Content-Type: application/json; charset=utf-8
Expires: -1
Vary: Accept-Encoding
Strict-Transport-Security: max-age=31536000; includeSubDomains
x-ms-request-id: fe2d1ebe-dab4-4f56-9af9-92346f9a354f
[...]

PR #853, opened 2019-10-23
use kudu https endpoint

weidongxu-microsoft:
test pass with both httpsOnly=true and httpsOnly=false

### DIFF
--- a/src/ResourceManagement/AppService/FunctionAppImpl.cs
+++ b/src/ResourceManagement/AppService/FunctionAppImpl.cs
@@ -363,7 +363,7 @@ namespace Microsoft.Azure.Management.AppService.Fluent
                 ServiceClientTracing.Enter(_invocationId, this, "Get", tracingParameters);
             }
             // Construct URL
-            var _baseUrl = string.Format("http://{0}", DefaultHostName().Replace("http://", "").Replace("https://", ""));
+            var _baseUrl = string.Format("https://{0}", DefaultHostName().Replace("http://", "").Replace("https://", ""));
             var _url = new System.Uri(new System.Uri(_baseUrl + (_baseUrl.EndsWith("/") ? "" : "/")), "admin/functions/{name}/keys").ToString();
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-net/issues/852

src/ResourceManagement/AppService/FunctionAppImpl.cs
```diff
@@ -363,7 +363,7 @@ public async Task<string> GetMasterKeyAsync(CancellationToken cancellationToken
                 ServiceClientTracing.Enter(_invocationId, this, "Get", tracingParameters);
             }
             // Construct URL
-            var _baseUrl = string.Format("http://{0}", DefaultHostName().Replace("http://", "").Replace("https://", ""));
```
yaohaizh: This changes the original implementation details by switch using https to https endpoint? Why original use http for https and what is the reason to change?

weidongxu-microsoft: It uses kudu. "http" will not work if function setting with httpsOnly=true. "https" will work for either httpsOnly=false or httpsOnly=true

yaohaizh: If this case, if httpsOnly=false, it will always use https, which change the meaning of API. Need conditional based on the httpsOnly settings.

weidongxu-microsoft: "https" should work for both. Kudu would suggest https even if you access http when httpsOnly=false.

```
> curl http://fa3-weidxu.scm.azurewebsites.net/
<head><title>Document Moved</title></head>
<body><h1>Object Moved</h1>This document may be found <a HREF="https://fa3-weidxu.scm.azurewebsites.net/">here</a></body>
```

I can do switch based on httpsOnly config, but since this request actually have Auth token in it, it would be better in HTTPS for all.

Issue #874, opened 2019-11-04
[BUG] Region member typo: Region.GovernmnetUSIowa


MiroslavVanicky:
Hi,

there is a typo in region name in Microsoft.Azure.Management.ResourceManager.Fluent.Core.Region: GovernmnetUSIowa.

yaohaizh:
@MiroslavVanicky thanks for this!

PR #883, opened 2019-11-08
Fix obsolete region field.

yaohaizh:


yaohaizh:
https://github.com/Azure/azure-libraries-for-net/issues/874

PR #884, opened 2019-11-08
Policy: generate package-policy-2019-06

xccc-msft:


src/ResourceManagement/ResourceManager/Domain/PolicyDefinition/Definition/IDefinition.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// The stage of the PolicyDefinition definition allowing to specify policy rule.
+    /// </summary>
+    public interface IWithPolicyRule
+    {
+        /// <summary>
+        /// Specifies the policy rule of PolicyDefinition.
+        /// </summary>
+        /// <param name="policyRule">The policy rule of the PolicyDefinition.</param>
+        /// <return>The next stage of the definition.</return>
+        IWithCreate WithPolicyRule(Object policyRule);
```
weidongxu-microsoft: Any idea why we would need to support a Object as input?

xccc-msft: This is how it gets defined in [definition swagger](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Authorization/stable/2019-06-01/policyDefinitions.json#L438)

weidongxu-microsoft: Do you know what object this is? If you do not know, do you think SDK user would know?

xccc-msft: Good catch. We could ask this question to [service side](https://docs.microsoft.com/en-us/rest/api/resources/policydefinitions/createorupdate)

yaohaizh: What the value from the Java side?

xccc-msft: In portal, the policy rule is json format. So it makes sense to have object parameters here. We have implemented object and json string both java and .net sides.

yaohaizh: Please document this find on the API docs.

PR #885, opened 2019-11-12
Authentication: support using credentials from Azure CLI automatically

xccc-msft:


src/ResourceManagement/ResourceManager/Authentication/AzureCliCredentials.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    using Newtonsoft.Json;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+
+    public class AzureCliCredentials
+    {
+        private static readonly string userProfile = "%USERPROFILE%";
```
weidongxu-microsoft: Are you sure it works both for windows and linux (and maybe mac)?

We got dotnet core that is cross platform.

src/ResourceManagement/ResourceManager/Authentication/AzureCliCredentials.cs
```diff
@@ -0,0 +1,84 @@
...
+                IEnumerable<AzureCliToken> tokens = JsonConvert.DeserializeObject<IEnumerable<AzureCliToken>>(File.ReadAllText(accessTokensPath));
+
+                while(wrapper == null || tokens == null || !tokens.Any() || wrapper.Subscriptions == null || !wrapper.Subscriptions.Any())
+                {
+                    Console.Error.WriteLine("Please login in Azure CLI and press any key to continue after you've successfully logged in.");
+                    Console.In.Read();
+
+                    wrapper = JsonConvert.DeserializeObject<AzureCliSubscriptionWrapper>(File.ReadAllText(azureProfilePath));
+                    tokens = JsonConvert.DeserializeObject<IEnumerable<AzureCliToken>>(File.ReadAllText(accessTokensPath));
+                }
```
weidongxu-microsoft: Move to private method if possible. "Don't repeat yourself", minimize duplicate code.

src/ResourceManagement/ResourceManager/Authentication/AzureCliSubscription.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Management.ResourceManager.Fluent.Authentication
+{
+    using Newtonsoft.Json;
+    using System;
+    using System.Collections.Generic;
+
+    public class AzureCliSubscription
```
weidongxu-microsoft: Does these need to be public? IMO we would not want user to see token easily.

PR #886, opened 2019-11-12
Allow specifying custom TokenCache

aloneguid:
This library doesn't allow specifying custom `TokenCache`, always defaulting to `TokenCache.Default`. This is OK when working with non-expiring credentials (i.e. service principal) but doesn't work when tokens need to be persisted.

### DIFF
--- a/src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
+++ b/src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
@@ -55,6 +55,11 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Authentication
 
         public AzureEnvironment Environment { get; private set; }
 
+        /// <summary>
+        /// Token cache to use with authentication, defaults to <see cref="TokenCache.DefaultShared"/>
+        /// </summary>
+        public TokenCache TokenCache { get; set; } = TokenCache.DefaultShared;
+
 #if NET45
         public AzureCredentials(UserLoginInformation userLoginInformation, string tenantId, AzureEnvironment environment)
             : this(tenantId, environment)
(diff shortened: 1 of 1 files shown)


yaohaizh:
@aloneguid Thanks for your contribution!

weidongxu-microsoft:
I think it looks good.

The DefaultShared would cache the token in memory.

@aloneguid Would you share a bit of your use case? Do you persist the token for later use, or for parellel use across process?

eosfor:
I wonder if this could help to reuse the token cache from an external context. for example, I'm making a PowerShell cmdlet using Fluent SDK. I want to use this cmdlet along with other cmdlets provided by [AZ module](https://docs.microsoft.com/en-us/powershell/azure/new-azureps-module-az?view=azps-3.8.0). In theory, I could try to reuse the AZ module token cache after it is authenticated, so I do not need to authenticate twice, once for the AZ module, and the second time for my Fluent cmdlet.

Or there is another way to reuse the cache?

@weidongxu-microsoft, what do you think?

weidongxu-microsoft:
@eosfor

I think the rationale underlying not sharing the token credential is mostly for security reason. The authenticate itself is not costly (maybe 1 additional request per hour?).

The `AzureCredentials` class itself is public. So it might be easier to extend the class and override `ProcessHttpRequestAsync`?

qub1n:
My rationale behind it is that we have multitenant application. The app is used by amny users which access different tenants and different subscriptions based on roles and accounts.

Thats why the app use multiple credentials to access multiple tenants/subscription.

Currently if there is only one shared token cache, we have to solve security boundary by application logic which is quite britle.

I would prefer to have security boundary by architecture a have token cache by user role or something similar.

weidongxu-microsoft:
@qub1n

So this PR will solve your use case? If so, I will merge it.

eosfor:
Hi @weidongxu-microsoft , basically my problem is [here](https://github.com/Azure/azure-libraries-for-net/issues/1075). I tried to overcome this by sharing AZ cache, but there are other changes necessary to make this all work. Seems like there are conflicting Auth libraries used in the AZ find Fluent, so they cannot be easily loaded into the pwsh process and some other things. But this PR will be just a part of the solution I think. But maybe, together with AZ module team you could find some other way to make it.

PR #887, opened 2019-11-13
feat: only run pipeline in master and pr

ChenTanyi:


.azure-pipelines/fluentNetSdk.yml
```diff
@@ -1,3 +1,6 @@
+trigger:
+  - master
```
weidongxu-microsoft: should be this?

trigger:
  branches:
    include:
      - master

ChenTanyi: this can work in https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#branches
and like java https://github.com/Azure/azure-libraries-for-java/blob/master/.azure-pipelines/fluentJavaMgmt.yml#L4

Issue #518, opened 2018-11-05
Traffic Manager needs way to configure IntervalInSeconds, ToleratedNumberOfFailures and TimeoutInSeconds


TeaTable:
Currently in the library I don't find a way to configure Traffic Manager monitor with intervalInSeconds, toleratedNumberOfFailures or timeoutInSeconds:

"monitorConfig": {
      "profileMonitorStatus": "Inactive",
      "protocol": "HTTP",
      "port": 80,
      "path": "/testpath.aspx",
      "intervalInSeconds": 30,
      "toleratedNumberOfFailures": 3,
      "timeoutInSeconds": 10
    }

The available interfaces I can find are:

Microsoft.Azure.Management.TrafficManager.Fluent.TrafficManagerProfile.Definition.IWithCreate WithHttpMonitoring();

Microsoft.Azure.Management.TrafficManager.Fluent.TrafficManagerProfile.Definition.IWithCreate WithHttpMonitoring(int port, string path);

Microsoft.Azure.Management.TrafficManager.Fluent.TrafficManagerProfile.Definition.IWithCreate WithHttpsMonitoring();

Microsoft.Azure.Management.TrafficManager.Fluent.TrafficManagerProfile.Definition.IWithCreate WithHttpsMonitoring(int port, string path);

jsukhabut:
This is becoming important for us now to use the Fast Failover feature.

https://azure.microsoft.com/en-us/updates/fast-failover-and-tcp-probing-in-azure-traffic-manager/

jsukhabut:
@hovsepm  I am currently working on adding this to the library.

I am also going to add enabled/disable Traffic View feature.

yaohaizh:
@weidongxu-microsoft  please take a look at this issue after the traffic manager is updated.

weidongxu-microsoft:
Fixed in 1.29. To be released on Dec. 9th.

Sample code:

```
                    var trafficManagerProfile = azure.TrafficManagerProfiles.Define(profileName)
                            ...
                            .WithHttpsMonitoring()
                            .WithFastFailover(10, 5)
```

PR #902, opened 2019-11-21
support fast failover setting to traffic manager

weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-net/issues/518


src/ResourceManagement/TrafficManager/TrafficManagerProfileImpl.cs
```diff
@@ -253,6 +268,14 @@ public TrafficManagerProfileImpl WithHttpsMonitoring(int port, string path)
             return this;
         }
 
+        public TrafficManagerProfileImpl WithFastFailover(long? intervalInSeconds, long? timeoutInSeconds, long? toleratedNumberOfFailures = default(long?))
+        {
+            Inner.MonitorConfig.IntervalInSeconds = intervalInSeconds;
```
ChenTanyi: I think parameter does not need `?`, and should check if `MonitorConfig` is null

weidongxu-microsoft: User is able to set null to use default value on service side. The interface does not encourage that (hence only the last param is optional), but I would still let user to set null if they really want that.

I will create Inner.MonitorConfig if null.

PR #911, opened 2019-11-29
Compute: update 2019 07 01

ChenTanyi:


ChenTanyi:
https://github.com/Azure/azure-libraries-for-java/issues/903

src/ResourceManagement/Compute/Domain/GalleryImageVersion/Definition/IDefinition.cs
```diff
@@ -170,13 +171,15 @@ public interface IWithSource  :
         /// </summary>
         /// <param name="customImageId">The ARM id of the custom image.</param>
         /// <return>The next definition stage.</return>
+        [Obsolete("WithSourceCustomImage is deprecated, source image was removed from REST API")]
```
yaohaizh: Do we have alternative methods for this API?

ChenTanyi: I find it may be move into `GalleryImageVersionStorageProfile`, I would change this implementation

ChenTanyi: fix, it would behave same as previous

src/ResourceManagement/Compute/Domain/InterfaceImpl/DiskImpl.cs
```diff
@@ -432,9 +430,39 @@ Disk.Definition.IWithCreateAndSize Disk.Definition.IWithWindowsDiskSource.WithWi
...
+        Disk.Definition.IWithStorageAccount Disk.Definition.IWithWindowsDiskSource.WithWindowsFromVhd(string vhdUrl)
         {
             return this.WithWindowsFromVhd(vhdUrl);
         }
+
+        /// <summary>
+        /// Specifies the storage account id.
+        /// </summary>
+        /// <param name="storageAccountId">The storage account id</param>
+        /// <returns>The next stage of the definition</returns>
```
weidongxu-microsoft: Check whether you need to add "."

ChenTanyi: The "." is just needed in summary, not for parameters

Issue #547, opened 2018-12-05
AKS - How to create virtual nodes when provisioning cluster


andtii:
We need to be able to create clusters with Virtual Node support enabled. Cant find how to do it!

yaohaizh:
Could you investigate whether we have support this feature?

andtii:
@ChenTanyi  @yaohaizh Any progress here?

ChenTanyi:
@andtii I haven't investigated it. Maybe I'll try to investigate it next sprint.

ChenTanyi:
@andtii the creation for virtual nodes uses the aciConnectorLinux add on, it could be written as:
```csharp

            var config = new Dictionary<string, string>();
            config.Add("SubnetName", virtualnetName);
            var addonProfile = new ManagedClusterAddonProfile(true, config);

            var addonMap = new Dictionary<string, ManagedClusterAddonProfile>();
            addonMap.Add("aciConnectorLinux", addonProfile);

            var virtualNetwork = azure.Networks.Define(networkName)
                .WithRegion(region)
                .WithExistingResourceGroup(rgName)
                .WithAddressSpace("10.0.0.0/8")
                .WithSubnet(akssubnetName, "10.240.0.0/16")
                .WithSubnet(virtualnetName, "10.241.0.0/16")
                .Create();

            var aks = azure.KubernetesClusters.Define(aksName)
                .WithRegion(region)
                .WithExistingResourceGroup(rgName)
                .WithLatestVersion()
                .WithRootUsername(username)
                .WithSshKey(sshKey)
                .WithServicePrincipalClientId(credentials.ClientId)
                .WithServicePrincipalSecret(clientSecret)
                .DefineAgentPool(agentPoolName)
                    .WithVirtualMachineSize(Microsoft.Azure.Management.ContainerService.Fluent.Models.ContainerServiceVMSizeTypes.StandardD2V2)
                    .WithAgentPoolVirtualMachineCount(1)
                    .WithVirtualNetwork(virtualNetwork.Id, akssubnetName)
                    .Attach()
                .WithDnsPrefix(dnsPrefixName)
                .DefineNetworkProfile
                    .WithNetworkPlugin(NetworkPlugin.Azure)
                    .Attach()
                .WithAddOnProfiles(addonMap)
                .Apply();
```
we may discuss about whether we should implement it as fluent interface.

yaohaizh:
Fixed in PR#914.

PR #914, opened 2019-12-02
Container Service: add virtual node in kubernetes cluster

ChenTanyi:


ChenTanyi:
https://github.com/Azure/azure-libraries-for-net/issues/547

src/ResourceManagement/ContainerService/Domain/KubernetesCluster/Definition/IDefinition.cs
```diff
@@ -357,6 +359,21 @@ public interface IWithAddOnProfilesBeta  :
...
+
+
+    /// <summary>
+    /// The stage of the Kubernetes cluster definition allowing to enable virtual node.
+    /// </summary>
+    public interface IWithVirtualNode :
+        Microsoft.Azure.Management.ResourceManager.Fluent.Core.IBeta
+    {
+        /// <summary>
+        /// Create a virtual node with ACI.
```
weidongxu-microsoft: `Creates ...`

Similar for update stage.

src/ResourceManagement/ContainerService/Domain/KubernetesCluster/Definition/IDefinition.cs
```diff
@@ -357,6 +359,21 @@ public interface IWithAddOnProfilesBeta  :
...
+        /// </summary>
+        /// <param name="subnetName">The subnet in AKS network to create virtual node.</param>
+        /// <returns>The next stage of the definition.</returns>
+        Microsoft.Azure.Management.ContainerService.Fluent.KubernetesCluster.Definition.IWithCreate WithVirtualNode(string subnetName);
```
weidongxu-microsoft: It might actually conflix with `IWithAddOnProfiles` stage, which set the whole `Inner.AddonProfiles`. The least to do is docstring mentioning this should be called before `WithAddOnProfiles`.

ChenTanyi:
Done, I'll try to make VS auto format in one single commit next.

Issue #915, opened 2019-12-03
Thumbprint on the portal seems incorrect after calling DefineCertificateCredential


huajunzhao:
**Describe the bug**

I use the sample code like [this](https://github.com/Azure/azure-libraries-for-net/blob/5f85a978aaa0340e5ef9f9747d3fefe052ad2a2b/Tests/Fluent.Tests/GraphRbac/ApplicationsTests.cs) to add a certificate into a AAD application, after code executed, when looking at the portal. I see the thumbprint is "63657274", can you please check if this is bug?

***Exception or Stack Trace***
Add the exception log and stack trace if available
No

**To Reproduce**

***Code Snippet***
Add the code snippet that causes the issue.

// certificate credentials definition
                    aadApptoCreate.DefineCertificateCredential("cert")
                                .WithAsymmetricX509Certificate()
                                .WithPublicKey(System.IO.File.ReadAllBytes(cert))
                                .WithDuration(TimeSpan.FromDays(100))
                                .Attach();
                }

**Expected behavior**

The thumbprint on the portal should be fixed.

Additional, i want to know how to change cert to anther one when updating the existing Aad application? Please share the code snippet.

**Screenshots**
![image](https://user-images.githubusercontent.com/3183325/70034376-0646f480-15ec-11ea-8965-6a3b48915695.png)

**Setup (please complete the following information):**
 - IDE : [e.g. IntelliJ]
 - Version of the Library used

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

yaohaizh:
Thanks for reporting the issue. We are looking into the issue. @ChenTanyi

ChenTanyi:
@huajunzhao I have test about it, it just display different from what was uploaded by portal (use different key), but the values are same.
certificate can not be changed, you can only create a new one.

PR #930, opened 2019-12-13
Graph RBAC: add thumbprint in certificate definition

ChenTanyi:
fix https://github.com/Azure/azure-libraries-for-net/issues/915

### DIFF
--- a/src/ResourceManagement/Graph.RBAC/CertificateCredentialImpl.cs
+++ b/src/ResourceManagement/Graph.RBAC/CertificateCredentialImpl.cs
@@ -140,11 +140,11 @@ namespace Microsoft.Azure.Management.Graph.RBAC.Fluent
             }
         }
 
-        internal CertificateCredentialImpl(string name, IHasCredential<T> parent)
+        internal CertificateCredentialImpl(string name, IHasCredential<T> parent, string customKeyIdentifier)
             : base(name, new KeyCredential()
             {
                 Usage = "Verify",
-                CustomKeyIdentifier = System.Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes(name)),
+                CustomKeyIdentifier = customKeyIdentifier,
                 StartDate = DateTime.Now,
                 EndDate = DateTime.Now.AddYears(1)
             })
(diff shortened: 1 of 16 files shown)


src/ResourceManagement/Graph.RBAC/Domain/ICertificateCredential.cs
```diff
@@ -13,5 +13,11 @@ public interface ICertificateCredential :
         Microsoft.Azure.Management.Graph.RBAC.Fluent.ICredential,
         Microsoft.Azure.Management.ResourceManager.Fluent.Core.IHasInner<Models.KeyCredential>
     {
+        /// <summary>
+        /// Custom Key Identifier. If the credential is defined by .NET SDK with name,
+        /// it would be the base64 encoding of name. If it is set by other tools, it would
+        /// be that value. Otherwise, it would usually be the thrumbprint of cretificate.
```
weidongxu-microsoft: spelling: thrumbprint and cretificate

ChenTanyi: done

ChenTanyi:
@weidongxu-microsoft @yaohaizh I add the deleting of these credential, you could review again.

src/ResourceManagement/Graph.RBAC/CertificateCredentialImpl.cs
```diff
@@ -140,11 +140,12 @@ internal CertificateCredentialImpl(KeyCredential keyCredential)
             }
         }
 
-        internal CertificateCredentialImpl(string name, IHasCredential<T> parent, string customKeyIdentifier)
+        internal CertificateCredentialImpl(string name, IHasCredential<T> parent, string customKeyIdentifier, Guid? keyId)
             : base(name, new KeyCredential()
             {
                 Usage = "Verify",
                 CustomKeyIdentifier = customKeyIdentifier,
+                KeyId = (keyId ?? Guid.NewGuid()).ToString(),
```
weidongxu-microsoft: So previous KeyId is null. And service would generate something for it upon PUT?

ChenTanyi: Yeah, the service will generate a UUID as KeyID, if it is null. The azure-cli generate keyId by itself, so I think this behavior would not be changed. https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/role/custom.py#L1604

Issue #935, opened 2019-12-19
[FEATURE REQ] Support change windows web app runtime


weidongxu-microsoft:
**Is your feature request related to a problem? Please describe.**

Runtime of Windows web app is configured via /config/metadata, which is not exposed in Fluent. And the config is obscure without proper constant definition.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Description Added
- [ ] Expected solution specified

weidongxu-microsoft:
1.30 (scheduled to be released on 9th, Jan.) will support e.g. WithRuntimeStack(WebAppRuntimeStack.NETCore) to update web app runtime stack for windows OS.
It need to be used in couple with existing WithJavaVersion+WithWebContainer / WithPythonVersion / WithPhpVersion etc. to set version.

PR #945, opened 2019-12-27
support change runtime stack for web app on windows OS

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/935

### DIFF
--- a/src/ResourceManagement/AppService/WebAppRuntimeStack.cs
+++ b/src/ResourceManagement/AppService/WebAppRuntimeStack.cs
@@ -1,13 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-/// <summary>
-/// Defines web app runtime stack on Windows operating system.
-/// </summary>
 namespace Microsoft.Azure.Management.AppService.Fluent
 {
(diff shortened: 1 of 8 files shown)


src/ResourceManagement/AppService/WebAppRuntimeStack.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Management.AppService.Fluent
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Defines web app runtime stack on Windows operating system.
+    /// </summary>
+    public class WebAppRuntimeStack
```
yaohaizh: 1. Should use ExpandableEnum? 
2. Since this if for Windows, will there be WebAppRuntimeStack for linux?

weidongxu-microsoft: Linux OS is the class RuntimeStack.cs
https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/AppService/RuntimeStack.cs

This class is done similar to RuntimeStack. It does not use Enum for similar reason. We might later put Version or other property in.

Currently this would used in couple with existing e.g. WithPhpVersion etc.

Issue #928, opened 2019-12-12
[BUG] RouteTables is not exposed properly


weidongxu-microsoft:
**Describe the bug**

RouteTables should be accessed via `Azure.RouteTables`.

Java got it right at `Azure.routeTables()`.

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**

***Code Snippet***
Add the code snippet that causes the issue.

**Expected behavior**

**Screenshots**

**Setup (please complete the following information):**
 - IDE : [e.g. IntelliJ]
 - Version of the Library used

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

xccc-msft:
Close this as released in v1.30.0

PR #950, opened 2020-01-07
Network: support private link service and Azure firewall

xccc-msft:


src/ResourceManagement/Network/AzureFirewallApplicationRuleCollectionImpl.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        IReadOnlyList<AzureFirewallApplicationRule> IAzureFirewallApplicationRuleCollection.Rules
+        {
+            get
+            {
+                List<AzureFirewallApplicationRule> rules = new List<AzureFirewallApplicationRule>();
+                if (Inner.Rules != null)
+                {
+                    rules = new List<AzureFirewallApplicationRule>(Inner.Rules);
```
yaohaizh: If this conditional path, the list is new twice. Should avoid.

src/ResourceManagement/Network/PrivateLinkServiceImpl.cs
```diff
@@ -0,0 +1,885 @@
...
+
+        public IReadOnlyList<ILoadBalancerFrontend> LoadBalancerFrontendIpConfigurations
+        {
+            get
+            {
+                List<ILoadBalancerFrontend> loadBalancerFrontends = new List<ILoadBalancerFrontend>();
+                Dictionary<string, LoadBalancerInner> loadBalancers = new Dictionary<string, LoadBalancerInner>();
+                if (Inner.LoadBalancerFrontendIpConfigurations != null)
+                {
+                    LoadBalancerInner loadBalancer;
```
yaohaizh: Keep the variable definition insider the loop

src/ResourceManagement/Network/PrivateLinkServiceImpl.cs
```diff
@@ -0,0 +1,885 @@
...
+        }
+
+        public IReadOnlyList<string> Fqdns
+        {
+            get
+            {
+                List<string> fqdns = new List<string>();
+                if (Inner.Fqdns != null)
+                {
+                    fqdns = new List<string>(fqdns);
```
yaohaizh: Avoid unnecessary two object creation

weidongxu-microsoft:
Could you also handle https://github.com/Azure/azure-libraries-for-net/issues/928?

src/ResourceManagement/Network/AzureFirewallImpl.cs
```diff
@@ -0,0 +1,1298 @@
...
+
+        private AzureFirewallImpl WithAzfwVnetSkuName()
+        {
+            if (Inner.Sku == null)
+            {
+                Inner.Sku = new AzureFirewallSku();
+            }
+            Inner.Sku.Name = AzureFirewallSkuName.AZFWVNet;
+            return this;
+        }
```
weidongxu-microsoft: Should we name it e.g WithAzureFirewallHubSku (Azfw -> full name, drop the Name since we know it is SKU)?
I am not sure about the background on VNet or Hub. So this suggestion might be wrong.
@yaohaizh

src/ResourceManagement/Network/AzureFirewallImpl.cs
```diff
@@ -0,0 +1,1298 @@
...
+
+        private AzureFirewallImpl WithFullQualifiedDomainNameTagsInApplicationRule(IList<string> tags)
+        {
+            applicationRule.FqdnTags = tags;
+            return this;
+        }
+
+        private AzureFirewallImpl WithHttpProtocolInApplicationRule(int? port)
```
weidongxu-microsoft: What happens when port is null?

src/ResourceManagement/Network/Domain/IAzureFirewallIpConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Gets reference of the PublicIP resource.
+        /// </summary>
+        string PublicIPAddressId { get; }
+
+        /// <summary>
+        /// Gets the provisioning state of the Azure firewall IP configuration
+        /// resource. Possible values include: 'Succeeded', 'Updating',
+        /// 'Deleting', 'Failed'
```
weidongxu-microsoft: end with "."

src/ResourceManagement/Network/AzureFirewallImpl.cs
xccc-msft: As per the [Azure Firewall Application Rule Protocol](https://docs.microsoft.com/en-us/rest/api/firewall/azurefirewalls/createorupdate#azurefirewallapplicationruleprotocol), the port is optional parameter. In the interface, I placed port as optional input arg.

src/ResourceManagement/Network/AzureFirewallImpl.cs
```diff
@@ -0,0 +1,1298 @@
...
+            get
+            {
+                ReadOnlyDictionary<string, string> additionalProps;
+                if (Inner.AdditionalProperties != null)
+                {
+                    additionalProps = new ReadOnlyDictionary<string, string>(Inner.AdditionalProperties);
+                }
+                else
+                {
+                    additionalProps = new ReadOnlyDictionary<string, string>(new Dictionary<string, string>());
```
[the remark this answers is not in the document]
xccc-msft: The [constructor of ReadOnlyDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.readonlydictionary-2?view=netframework-4.8#constructors) only support this kind of param.

src/ResourceManagement/Network/PrivateLinkServicesImpl.cs
```diff
@@ -0,0 +1,231 @@
...
+                async (cancellation) => await Inner.ListAsync(resourceGroupName, cancellation),
+                Inner.ListNextAsync,
+                WrapModel,
+                loadAllPages,
+                cancellationToken);
+        }
+
+        public IEnumerable<IPrivateLinkService> ListBySubscription()
+        {
+            throw new System.NotImplementedException();
```
weidongxu-microsoft: ???

PR #970, opened 2020-02-05
Resources: set ApiVersion as optional parameter

xccc-msft:


src/ResourceManagement/ResourceManager/GenericResource/GenericResourceImpl.cs
```diff
@@ -143,9 +144,10 @@ public GenericResource.Definition.IWithPlan WithProviderNamespace(string resourc
             return this;
         }
 
-        public IWithCreate WithApiVersion(string apiVersion)
+        public IWithCreate WithApiVersion(string apiVersion = default(string))
         {
-            this.apiVersion = apiVersion;
+            this.apiVersion = apiVersion ?? ResourceUtils.ApiVersionFromResourceProviderAndType(resourceProviderNamespace, resourceType);
```
tomkerkhove: Just a thought - Now that we have the list of supported versions, could we validate the specified api version to see if it's supported or not?

xccc-msft: The list of supported versions come from the clients of current SDK. It could be used when users have no info about the api versions and such type of resource gets supported in our SDK. But there are wide range of resources/api versions in Azure. It's hard to validate specified api version now.

tomkerkhove:
Makes sense, thanks.
________________________________

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -158,13 +183,20 @@ public new IGenericResource GetById(string id)
 
         public async new Task<IGenericResource> GetByIdAsync(string id, CancellationToken cancellationToken = default(CancellationToken))
         {
+            string apiVersion = ResourceUtils.DefaultApiVersionFromResourceId(
+                id, 
+                await Manager.Providers.GetByNameAsync(ResourceUtils.ResourceProviderFromResourceId(id), cancellationToken));
```
adamconnelly: Am I right in thinking that `IProviders.GetByNameAsync()` makes a web request? If so, my one concern about this approach is that it's going to change this method from making a single request for the resource to two requests:

1. To get the `IProvider` instance.
2. To actually get the resource using the ApiVersion we've just looked up.

Would it be worth introducing some kind of caching to make sure that you don't have to lookup the version / the `IProvider` every single time? Otherwise I think it might end up adding quite a lot of overhead to this request, as well as using up more of the user's API rate limit.

src/ResourceManagement/ResourceManager/Core/ResourceUtils.cs
```diff
@@ -144,5 +144,38 @@ public static AzureEnvironment ExtractAzureEnvironment(RestClient restClient)
             return environment;
         }
 
+        public static string ResourceTypeForApiVersion(string id)
+        {
+            string type = ResourceTypeFromResourceId(id);
+            string parent = ParentResourcePathFromResourceId(id);
+            while (parent != null)
```
yaohaizh: Using String.IsNullOrEmpty for null check of string value.

src/ResourceManagement/ResourceManager/Core/ResourceUtils.cs
```diff
@@ -144,5 +144,38 @@ public static AzureEnvironment ExtractAzureEnvironment(RestClient restClient)
...
+            string resourceType = ResourceTypeForApiVersion(id);
+            if (resourceType == null)
+            {
+                return null;
+            }
+            if (provider != null && provider.ResourceTypes != null)
+            {
+                foreach (var providerResourceType in provider.ResourceTypes)
+                {
+                    if (resourceType.Equals(providerResourceType.ResourceType)
```
yaohaizh: Use String comparisoin of OrdinalIgnoreCase

src/ResourceManagement/ResourceManager/Core/ResourceUtils.cs
weidongxu-microsoft: CoreUtils class

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -38,9 +38,20 @@ public bool CheckExistence(string resourceGroupName, string resourceProviderName
...
+            if (apiVersion == null)
+            {
+                string resourceId = ResourceUtils.ConstructResourceId(Manager.Inner.SubscriptionId, resourceGroupName, resourceProviderNamespace, resourceType, resourceName, parentResourcePath);
+                apiVersion = ResourceUtils.DefaultApiVersionFromResourceId(
+                    resourceId,
+                    await Manager.Providers.GetByNameAsync(resourceProviderNamespace, cancellationToken));
+                if (apiVersion == null)
+                {
+                    throw new ArgumentException(string.Format("{0} is not a supported resource type in provider {1}", resourceType, resourceProviderNamespace));
+                }
```
weidongxu-microsoft: Try to do it in a method, at least in this class. Seems copied multiple times.

src/ResourceManagement/ResourceManager/GenericResource/GenericResourceImpl.cs
```diff
@@ -97,6 +98,17 @@ public void prepareDefaultValue()
         public async override Task<IGenericResource> CreateResourceAsync(CancellationToken cancellationToken)
         {
             prepareDefaultValue();
+            if (apiVersion == null)
```
yaohaizh: https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=netframework-4.8

src/ResourceManagement/ResourceManager/GenericResource/GenericResourceImpl.cs
```diff
@@ -97,6 +98,17 @@ public void prepareDefaultValue()
...
+            {
+                string resourceId = ResourceUtils.ConstructResourceId(Manager.Inner.SubscriptionId, ResourceGroupName, resourceProviderNamespace, resourceType, Name, parentResourceId);
+                apiVersion = ResourceUtils.DefaultApiVersionFromResourceId(
+                    resourceId, 
+                    await Manager.Providers.GetByNameAsync(resourceProviderNamespace, cancellationToken));
+                if (apiVersion == null)
+                {
```
yaohaizh: This is not an argument exception. At least, argument is not passed as parameter here.

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -17,6 +17,8 @@ internal class GenericResourcesImpl :
         TopLevelModifiableResources<IGenericResource, GenericResourceImpl, GenericResourceInner, IResourcesOperations, IResourceManager>,
         IGenericResources
     {
+        internal static IDictionary<string, string> cachedApiVersions = new Dictionary<string, string>();
```
yaohaizh: ConcurrencyDictionary

src/ResourceManagement/ResourceManager/GenericResource/GenericResourceImpl.cs
xccc-msft: The argument exception will throw when it cannot find API version when using resource provider/resource type/resource id from method signature. I think it would be ArgumentException. Otherwise lets use InvalidOperationException?

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -293,5 +271,29 @@ public async Task<IPagedCollection<IGenericResource>> ListByTagAsync(string reso
...
+            else
+            {
+                apiVersionValue = ResourceUtils.DefaultApiVersionFromResourceId(
+                    id,
+                    resourceManager.Providers.GetByName(resourceProvider));
+                if (apiVersionValue == null)
+                {
+                    throw new ArgumentException(string.Format("{0} is not a supported resource type in provider {1}", resourceType, resourceProvider));
+                }
+                cachedApiVersions.Add(apiVersionKey, apiVersionValue);
```
weidongxu-microsoft: DO NOT call Add in concurrency (it throws exception when find key exists). Use TryAdd or AddOrUpdate.

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -293,5 +271,29 @@ public async Task<IPagedCollection<IGenericResource>> ListByTagAsync(string reso
...
+        internal static string GetApiVersion(string id, IResourceManager resourceManager)
+        {
+            string apiVersionValue;
+            string resourceProvider = ResourceUtils.ResourceProviderFromResourceId(id);
+            string resourceType = ResourceUtils.ResourceTypeForApiVersion(id);
+            string apiVersionKey = resourceProvider + "/" + resourceType;
+            if (cachedApiVersions.ContainsKey(apiVersionKey))
+            {
+                apiVersionValue = cachedApiVersions[apiVersionKey];
+            }
```
weidongxu-microsoft: Use TryGetValue.

Normally, in concurrency, you want one method call to do the job, not 2.

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
```diff
@@ -293,5 +271,29 @@ public async Task<IPagedCollection<IGenericResource>> ListByTagAsync(string reso
                 Manager.Inner.Resources.ListByResourceGroupNextAsync,
                 WrapModel, loadAllPages, cancellationToken);
         }
+
+        internal static string GetApiVersion(string id, IResourceManager resourceManager)
```
weidongxu-microsoft: You should use async here?

src/ResourceManagement/ResourceManager/GenericResource/GenericResourcesImpl.cs
xccc-msft: Added cache for API versions.

Issue #750, opened 2019-07-26
[FEATURE REQ] Support outbound rules for standard sku load balancer


NatMarchand:
Hi,
Currently, it is not possible to create a load balancer standard sku with an outbound rule.
It it also not possible to create a load balancer without any rule (or just this outbound rule).

Ideally, I would like to declare something like this :
```
loadBalancer = await azure.LoadBalancers
    .Define(loadBalancerName)
    .WithRegion(resourceGroup.Region)
    .WithExistingResourceGroup(resourceGroup)
    .DefineOutboundRule("Outbound Rule")
        .WithProtocol(TransportProtocol.Tcp)
        .FromBackend(backend)
        .ToFrontend(frontend)
        .Attach()
    .DefinePublicFrontend(frontend)
        .WithExistingPublicIPAddress(ip)
        .Attach()
    .DefineBackend(backend)
        .Attach()
    .WithSku(LoadBalancerSkuType.Standard)
    .CreateAsync();
```

loarabia:
Thanks for reporting this issue. I've tagged it to get it routed to the right team.

yaohaizh:
We plan support for Feb. release.

weidongxu-microsoft:
Feature supported in 1.31.

Sample code:

```
                    ILoadBalancer loadBalancer = azure.LoadBalancers.Define(lbName)
                        .WithRegion(resourceGroup.Region)
                        .WithExistingResourceGroup(resourceGroup)
                        // Define outbound rule
                        .DefineOutboundRule(outboundRuleName1)
                            .WithProtocol(LoadBalancerOutboundRuleProtocol.Tcp)
                            .FromBackend(backendPoolName1)
                            .ToFrontend(frontEndName1)
                            .WithEnableTcpReset()
                            .WithIdleTimeoutInMinutes(10)
                            .Attach()
                        // Outbound rule need to use static public IP address with Standard SKU
                        .DefinePublicFrontend(frontEndName1)
                            .WithNewPublicIPAddress(azure.PublicIPAddresses.Define(publicIpName1).WithRegion(resourceGroup.Region).WithExistingResourceGroup(resourceGroup).WithSku(PublicIPSkuType.Standard).WithStaticIP())
                            .Attach()
                        .WithSku(LoadBalancerSkuType.Standard)
                        .Create();
```

PR #973, opened 2020-02-06
feature support load balancer outbound rule

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/750

### DIFF
--- a/src/ResourceManagement/Network/Domain/ILoadBalancerOutboundRule.cs
+++ b/src/ResourceManagement/Network/Domain/ILoadBalancerOutboundRule.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.Management.Network.Fluent
     using System.Collections.Generic;
 
     /// <summary>
-    /// A client-side representation of a outbound rule of the load balancer.
+    /// A client-side representation of an outbound rule of the load balancer.
     /// </summary>
     public interface ILoadBalancerOutboundRule :
         IHasInner<OutboundRuleInner>,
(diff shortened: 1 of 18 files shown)


src/ResourceManagement/Network/Domain/ILoadBalancerOutboundRule.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+namespace Microsoft.Azure.Management.Network.Fluent
+{
+    using Microsoft.Azure.Management.Network.Fluent.Models;
+    using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A client-side representation of a outbound rule of the load balancer.
```
yaohaizh: a ==> an

src/ResourceManagement/Network/LoadBalancingRuleImpl.cs
```diff
@@ -339,5 +347,17 @@ internal LoadBalancingRuleImpl FromNewPublicIPAddress(string leafDnsLabel)
             Parent.WithNewPublicIPAddress(leafDnsLabel, frontendName);
             return FromFrontend(frontendName);
         }
+
+        internal LoadBalancingRuleImpl WithDisableOutboundSnat()
+        {
```
yaohaizh: If this two methods duplicated?

weidongxu-microsoft: The method without parameter is used in definition stage, when user not calling it, it is just as default true. The method with boolean parameter is used in update stage, user can change true to false, or false to true.

PR #976, opened 2020-02-10
Release note 1.31.0

ChenTanyi:


ChenTanyi:
@xccc-msft I didn't see it either.

xccc-msft:
@weidongxu-microsoft @ChenTanyi 
It can be found in [this line](https://github.com/Azure/azure-libraries-for-net/pull/970/files#diff-47fc0efd1d90f60e13b0a8430c0274c7L20-L21). Inside the impl, it will throw UnsupportedException.

weidongxu-microsoft:
I see...

```
        protected override Task<GenericResourceInner> GetInnerByGroupAsync(string groupName, string name, CancellationToken cancellation)
        {
            throw new NotSupportedException("Get just by resource group and name is not supported. Please use other overloads.");
        }
```

Issue #992, opened 2020-03-03
[BUG] Azure.AppServices.FunctionApps.ListAsync() doesn't work as expected C# SDK v1.31


weidongxu-microsoft:
ICM https://icm.ad.msft.net/imp/v3/incidents/details/177404609/home

weidongxu-microsoft:
Sync method from web app and function app, i.e. List / ListByResourceGroup is not filtering at all. Basically they return web app and function app, regardless whether it is called in WebApps or FunctionApps.

Async method from function app got a bug that Linux function app is no included in result.

PR #993, opened 2020-03-03
proper filter for webapp and function app

weidongxu-microsoft:
from sev3 ICM

fix https://github.com/Azure/azure-libraries-for-net/issues/992

### DIFF
--- a/src/ResourceManagement/AppService/FunctionAppsImpl.cs
+++ b/src/ResourceManagement/AppService/FunctionAppsImpl.cs
@@ -56,12 +56,12 @@ namespace Microsoft.Azure.Management.AppService.Fluent
                 Inner.ListNextAsync,
                 async (inner, cancellation) => await PopulateModelAsync(inner, cancellation),
                 loadAllPages, cancellationToken);
-            return PagedCollection<IFunctionApp, SiteInner>.CreateFromEnumerable(collection.Where(w => w.Inner.Kind != null && w.Inner.Kind.Split(new char[] { ',' }).Contains("functionapp")));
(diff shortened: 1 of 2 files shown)


src/ResourceManagement/AppService/FunctionAppsImpl.cs
```diff
@@ -58,7 +56,12 @@ public override async Task<IPagedCollection<IFunctionApp>> ListAsync(bool loadAl
                 Inner.ListNextAsync,
                 async (inner, cancellation) => await PopulateModelAsync(inner, cancellation),
                 loadAllPages, cancellationToken);
-            return PagedCollection<IFunctionApp, SiteInner>.CreateFromEnumerable(collection.Where(w => "functionapp".Equals(w.Inner.Kind, StringComparison.OrdinalIgnoreCase)));
+            return PagedCollection<IFunctionApp, SiteInner>.CreateFromEnumerable(collection.Where(w => w.Inner.Kind != null && w.Inner.Kind.Split(new char[] { ',' }).Contains("functionapp")));
```
ChenTanyi: This seems duplicate

xccc-msft: Not sure why we remove StringComparison.OrdinalIgnoreCase here

weidongxu-microsoft: Thanks for point this out. I should use `FilterFunctionApp` uniformly.

For ignoreCase, I will add `ToLower` before `Split` as protection. Thought it should not happen in live.

Issue #996, opened 2020-03-06
[BUG]Dead loop occurs when NextPageLink is not null but resource results are empty


jasonxdhu:
**Describe the bug**
In the method **AddCollection** of class **PagedCollection**, there is dead-loop when the resources returned from service are empty but next-page-link is not null.

It should continue invoking requests to get more results based-on next-page-link until it reaches to null. It cannot assume there are no more results because the current results are empty. The common practice is to follow the next-page-link until it reaches to null, which indicates no more results from server.

***Exception or Stack Trace***
No exception but dead-loop occurs. 

**To Reproduce**
Mock up a Page’s data returned from server to zero records but NextPageLink is not null.

***Code Snippet***
The code is in file src\ResourceManagement\ResourceManager\Core\PagedCollection.cs.

The existing code:
        private async Task AddCollection(IPage<InnerResourceT> currentPage, PagedCollection<IFluentResourceT, InnerResourceT> pagedCollection,
            CancellationToken cancellationToken)
        {
            if (currentPage != null && currentPage.Any())
            {
                pagedCollection.NextPageLink = currentPage.NextPageLink;
                var resources = await Task.WhenAll(currentPage.Select(async (inner) => await this.WrapModelAsyncDelegate(inner, cancellationToken)));
                ((List<IFluentResourceT>)pagedCollection.innerCollection).AddRange(resources);
            }
        }

This code will cause dead-loop in method LoadPageWithWrapModelAsync‘s do-while loop when currentPage.Any() is **false** since pagedCollection.NextPageLink is NOT updated with next one.

**Expected behavior**
        private async Task AddCollection(IPage<InnerResourceT> currentPage, PagedCollection<IFluentResourceT, InnerResourceT> pagedCollection,
            CancellationToken cancellationToken)
        {
            if (currentPage != null)
            {
                pagedCollection.NextPageLink = currentPage.NextPageLink;
                if (currentPage.Any())
                {
                    var resources = await Task.WhenAll(currentPage.Select(async (inner) => await this.WrapModelAsyncDelegate(inner, cancellationToken)));
                    ((List<IFluentResourceT>)pagedCollection.innerCollection).AddRange(resources);
                }
            }
        }
This will pass current page NextPageLink to pagedCollection’s one. It will continue making requests by following the next link until it reaches null and stop looping in method LoadPageWithWrapModelAsync‘s do-while loop.

**Setup (please complete the following information):**
 - OS: Windows 10
[...]

PR #999, opened 2020-03-06
fix stalling on PagedCollection, when value is empty but nextLink is valid

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/996

### DIFF
--- a/Tests/Fluent.Tests/PagedCollectionTests.cs
+++ b/Tests/Fluent.Tests/PagedCollectionTests.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.KeyVault.Models;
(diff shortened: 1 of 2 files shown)


yaohaizh:
@weidongxu-microsoft I'm okay with this change. But my second taking is that whether the KeyVault team break the protocol of the swagger spec for nextLinkPage? https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md#x-ms-pageable

weidongxu-microsoft:
@yaohaizh 

Keyvault actually have an empty value, with valid nextLink. I.e., they seems got one page which is empty, and still need to continue the paging.

My guess is they are doing some filtering internally on the page result before returning to us.

@jasonxdhu 

Would you please share more on this case?

yungezz:
agree with @yaohaizh , other sdks should meet similar issue: nextLInk valid while empty page content. Can we have a look at Keyvault swagger and file issue there if indeed a swagger issue?

PR #1000, opened 2020-03-09
Compute: support get gallery image version with replication status

xccc-msft:


src/ResourceManagement/Compute/Domain/IGalleryImageVersions.cs
```diff
@@ -82,5 +82,27 @@ public interface IGalleryImageVersions  :
...
         Task<IPagedCollection<Microsoft.Azure.Management.Compute.Fluent.IGalleryImageVersion>> ListByGalleryImageAsync(string resourceGroupName, string galleryName, string galleryImageName, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Retrieves information about a gallery image version.
+        /// </summary>
+        /// <param name="resourceGroupName">The name of the resource group.</param>
+        /// <param name="galleryName">The name of the gallery.</param>
+        /// <param name="galleryImageName">The name of the gallery image.</param>
+        /// <param name="galleryImageVersionName">The name of the gallery image version.</param>
+        /// <throws>IllegalArgumentException thrown if parameters fail the validation.</throws>
```
weidongxu-microsoft: I know it is copied... but this one is definitely not correct?

PR #1004, opened 2020-03-11
Compute: append storage account id when create option is import

xccc-msft:


src/ResourceManagement/Compute/Domain/Snapshot/Definition/IDefinition.cs
```diff
@@ -277,11 +277,11 @@ public interface IWithWindowsSnapshotSource
         Microsoft.Azure.Management.Compute.Fluent.Snapshot.Definition.IWithCreate WithWindowsFromSnapshot(ISnapshot sourceSnapshot);
 
         /// <summary>
-        /// Specifies the source specialized or generalized Windows OS VHD.
+        /// Specifies the source specialized or generalized Windows OS VHD and the storage account ID.
         /// </summary>
         /// <param name="vhdUrl">The source VHD URL.</param>
         /// <return>The next stage of the definition.</return>
```
yaohaizh: Need update comments

PR #1011, opened 2020-03-16
Network: support AzureFirewall update operation for rule collections

xccc-msft:


src/ResourceManagement/Network/AzureFirewallImpl.cs
```diff
@@ -457,37 +462,17 @@ protected async override Task<AzureFirewallInner> GetInnerAsync(CancellationToke
...
-                Inner.NetworkRuleCollections.Add(networkRuleCollection);
-                networkRuleCollection = null;
-            }
-            else if (ipConfiguration != null)
-            {
-                Inner.IpConfigurations.Add(ipConfiguration);
-                ipConfiguration = null;
-            }
-            return this;
+            return Attach();
```
yungezz: what's impact of remove existing appending rule action?

yungezz:
HI @xccc-msft why this change is important in net fluent now? is same behavior in java fluent now?

yaohaizh:
@yungezz this is requested by service team. On the other hand, the java(v1 & v2) lags behind for this support because Java doesn't support private DNS. Maybe, private DNS is a good service for you to ramp up

src/ResourceManagement/Network/AzureFirewallImpl.cs
xccc-msft: I just move this appending action into define method. This is to align with update operations which don't required appending action.

PR #1015, opened 2020-03-18
Resources: support api version when listing generic resources

### LINKED ISSUE #978: [BUG] IGenericResource.ApiVersion is always null
**Describe the bug**
When retrieving an `IGenericResource `instance from `genericResources.ListByResourceGroupAsync`  `ApiVersion` property is null.

**To Reproduce**
The resource already exists and when showing ARM template in Azure `ApiVersion` is filled.
```
var resources = await genericResources.ListByResourceGroupAsync(Config.ResourceGroup);
foreach (var resource in resources)
{
    // resource.ApiVersion is always null
    await resource.Update().WithApiVersion(resource.ApiVersion).WithTags(new DefaultTags()).ApplyAsync();
}
```
**Expected behavior**
`IGenericResource.ApiVersion` should be filled with the correct `ApiVersion` that was used when the resource was created or updated.

**Setup (please complete the following information):**
 - Microsoft.Azure.Management.Fluent 1.31.0
 - Microsoft.Azure.Management.ResourceManager.Fluent 1.31.0

xccc-msft:
fix #978

### DIFF
--- a/Tests/Fluent.Tests/ResourceManager/GenericResourcesTests.cs
+++ b/Tests/Fluent.Tests/ResourceManager/GenericResourcesTests.cs
@@ -139,6 +139,7 @@ namespace Fluent.Tests.ResourceManager
                                  where string.Equals(r.Name, resourceName, StringComparison.OrdinalIgnoreCase)
                                  select r).FirstOrDefault();
                     Assert.NotNull(found);
+                    Assert.NotNull(found.ApiVersion);
 
                     // Get
                     resource = genericResources.Get(rgName,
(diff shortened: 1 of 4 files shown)


xccc-msft:
> Looks OK. Though it might be a bit concern that we do an extra request only for some user case that need the apiVersion (maybe more cases will not need it).

For users who want to do some update operation, it's not extra request as they will need apiVersion. For others, they can avoid the extra request by not using resource.apiVersion().

weidongxu-microsoft:
Nice. It is lazy evaluation. So no cost if no use.

PR #1019, opened 2020-03-20
Resources: support TaskCanceledException handling for different .NET environment

### LINKED ISSUE #1002: [BUG] DNS record update failed with 504 after about 2min but no exception was thrown
**Describe the bug**
Sometimes ARM returns 504 Gateway Timeout after about 1-2min. However the library throws no exception so the caller is unable to know the operation actually fails.

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**
Use Fiddler to inject **1min delay and 504 response** for DNS A record update request:

image

Exported Fiddler rules:

```
<?xml version="1.0" encoding="utf-8"?>
<AutoResponder LastSave="2020-03-11T13:04:20.1331365+08:00" FiddlerVersion="4.6.20171.26113">
  <State Enabled="true" Fallthrough="true" UseLatency="false">
    <ResponseRule Match="EXACT:https://management.azure.com/subscriptions/x/resourceGroups/x/providers/Microsoft.Network/dnsZones/example.com/A/foobar?api-version=2018-03-01-preview" Action="*delay:60000" Enabled="true" />
    <ResponseRule Match="EXACT:https://management.azure.com/subscriptions/x/resourceGroups/x/providers/Microsoft.Network/dnsZones/example.com/A/foobar?api-version=2018-03-01-preview" Action="C:\Users\work\Desktop\Response.txt" Enabled="true" />
  </State>
</AutoResponder>
```

Response file content(Response.txt):
```
HTTP/1.1 504 Gateway Timeout
Content-Type: application/json; charset=utf-8

{"error":{"details":[],"code":"GatewayTimeout"}}
```

Run the code.

Here is the mock scenario:

image

***Code Snippet***

```csharp
var creds = new AzureCredentialsFactory().FromServicePrincipal(
    "x", "x", "x",
    AzureEnvironment.AzureGlobalCloud);
var restClient = RestClient.Configure()
    .WithEnvironment(creds.Environment)
    .WithCredentials(creds)
    .Build();

// Use Fiddler proxy
restClient.RootHttpHandler.Proxy = new WebProxy("http://localhost:8888", true);
restClient.RootHttpHandler.UseProxy = true;
restClient.RootHttpHandler.ServerCertificateCustomValidationCallback = (HttpRequestMessage r, X509Certificate2 c, X509Chain ch, SslPolicyErrors e) =>
{
    return true;
};

var azure = Azure.Authenticate(restClient, creds.TenantId)
    .WithSubscription("x");

var zone = azure.DnsZones.GetById("/subscriptions/x/resourceGroups/x/providers/Microsoft.Network/dnszones/example.com");
zone.Update()
    .DefineARecordSet("foobar")
    .WithIPv4Address("1.1.1.1")
    .Attach()
    .ApplyAsync().Wait();
```

**Expected behavior**
The library should throw an exception.

**Screenshots**

**Setup (please complete the following information):**
Windows
 - IDE : [e.g. IntelliJ]
 - Version of the Library used
1.31.1

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

xccc-msft:
Fix #1002 
This was caused by the difference from dotnet frameworks when getting TaskCanceledException. For example, `netcoreapp2.0` will set task as `Faulted` while `net461` will set as `Canceled`.

### DIFF
--- a/src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
+++ b/src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Core
                                 exceptions.Add(deleteTask.Exception);
                             }
                         }
-                        if (deleteTask.IsCanceled)
+                        else if (deleteTask.IsCanceled)
                         {
                             exceptions.Add(new TaskCanceledException());
                         }
--- a/src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
+++ b/src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
@@ -69,12 +69,15 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Core
                                 exceptions.Add(deleteTask.Exception);
                             }
                         }
+                        if (deleteTask.IsCanceled)
+                        {
+                            exceptions.Add(new TaskCanceledException());
+                        }
                         else
                         {
                             comitted.Add(res);
                             res.PendingOperation = PendingOperation.None;
-                            FluentModelTImpl val;
-                            this.collection.TryRemove(res.Name(), out val);
+                            this.collection.TryRemove(res.Name(), out FluentModelTImpl val);
                         }
                     },
                     cancellationToken,
@@ -91,8 +94,7 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Core
                     {
                         if (createTask.IsFaulted)
                         {
-                            FluentModelTImpl val;
-                            this.collection.TryRemove(res.Name(), out val);
+                            this.collection.TryRemove(res.Name(), out FluentModelTImpl val);
                             if (createTask.Exception.InnerException != null)
                             {
                                 exceptions.Add(createTask.Exception.InnerException);
@@ -101,6 +103,11 @@ namespace Microsoft.Azure.Management.ResourceManager.Fluent.Core
                             {
                                 exceptions.Add(createTask.Exception);
                             }
+                        }
+                        else if (createTask.IsCanceled)
+                        {
+                            this.collection.TryRemove(res.Name(), out FluentModelTImpl val);
+                            exceptions.Add(new TaskCanceledException());
                         }
                         else
                         {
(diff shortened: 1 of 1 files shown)


src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
```diff
@@ -69,12 +69,15 @@ public Task<List<FluentModelTImpl>> CommitAndGetAllAsync(CancellationToken cance
                                 exceptions.Add(deleteTask.Exception);
                             }
                         }
+                        if (deleteTask.IsCanceled)
```
yungezz: curious, why specific to deleteTask

src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
```diff
@@ -69,12 +69,15 @@ public Task<List<FluentModelTImpl>> CommitAndGetAllAsync(CancellationToken cance
...
+                        {
+                            exceptions.Add(new TaskCanceledException());
```
ArchangelSDY: What's the difference between `IsFaulted` branch and `IsCanceled` branch?

src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
xccc-msft: Before execution, we will label all pending child tasks with `ToDelete`/`ToCreate`/`ToUpdate`. And this situation may happen at any of the pending tasks, not only limited by one kind of them.

src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
xccc-msft: When getting 504 error, for `isFaulted`, the task would have taskCanceledException inside. But for `isCanceled`, the task would have exception as null.

src/ResourceManagement/ResourceManager/Core/ExternalChildResourceCollection.cs
ArchangelSDY: `else if` here?

xccc-msft: Updated. Thanks.

weidongxu-microsoft:
Just a comment, make sure it works as expected when it is really being cancelled by the CancellationToken.

PR #1023, opened 2020-03-24
adding MinTlsVersion

avantida:
To be able to review/report the security setting of app services, we'd like to be able to access the MinTlsVersion.


weidongxu-microsoft:
@avantida 
I think you need add it to the interface as well.
https://github.com/Azure/azure-libraries-for-net/blob/5528333fd6f0588e6ed80aad85b26e7af67d26e9/src/ResourceManagement/AppService/Domain/IWebAppBase.cs#L421

weidongxu-microsoft:
@avantida

Thanks for your contribution.

Look good to me. I will keep it open for one or two days for others to review, then I will merge the PR.

Next release is 1.33.0, expected on April 10.

Issue #1008, opened 2020-03-16
[FEATURE REQ] Cosmos DB remove write replication in simple definiton


ChenTanyi:
As https://github.com/Azure/azure-libraries-for-net/issues/1006#issuecomment-598631450 discussed about.

PR #1025, opened 2020-03-25
Cosmos: update interface

ChenTanyi:


src/ResourceManagement/CosmosDB/CosmosDBAccountImpl.cs
```diff
@@ -413,6 +413,11 @@ public CosmosDBAccountImpl WithWriteReplication(Region region, bool? isZoneRedun
             return this;
         }
 
+        public CosmosDBAccountImpl WithDefaultWriteReplication()
+        {
+            return this.WithWriteReplication(this.Region, null);
```
weidongxu-microsoft: Not sure whether this is needed or not. What happens if `WithWriteReplication` is skipped altogether?

ChenTanyi: Not very well to skip it. Since the withReadReplication will do the same thing to set locations. If sb. skip write and set a read replication, it will become write replication.

ChenTanyi:
fix #1024 #1008

PR #1031, opened 2020-03-31
Orpheus - skip shutdown

weidongxu-microsoft:
Try code change.


src/ResourceManagement/Compute/Domain/IVirtualMachine.cs
```diff
@@ -242,7 +242,14 @@ public interface IVirtualMachine :
...
         /// <return>A representation of the deferred computation of this call.</return>
-        Task PowerOffAsync(CancellationToken cancellationToken = default(CancellationToken), bool skipShutdown = false);
+        Task PowerOffAsync(CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Powers off (stops) the virtual machine asynchronously.
+        /// </summary>
+        /// <param name="skipShutdown">The parameter to request non-graceful VM shutdown. True value for this flag indicates non-graceful shutdown whereas false indicates otherwise.</param>
+        /// <return>A representation of the deferred computation of this call.</return>
+        Task PowerOffAsync(bool skipShutdown = false, CancellationToken cancellationToken = default(CancellationToken));
```
OrpheusQu: Just one suggestion, maybe don't give a default value to skipShutdown to differentiate two PowerOffAsync methods.

src/ResourceManagement/Compute/Domain/IVirtualMachineScaleSetVM.cs
```diff
@@ -272,7 +272,14 @@ public interface IVirtualMachineScaleSetVM :
...
         /// <return>A representation of the deferred computation of this call.</return>
-        Task PowerOffAsync(CancellationToken cancellationToken = default(CancellationToken), bool skipShutdown = false);
+        Task PowerOffAsync(CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Stops the virtual machine instance.
+        /// </summary>
+        /// <param name="skipShutdown">The parameter to request non-graceful VM shutdown. True value for this flag indicates non-graceful shutdown whereas false indicates otherwise.</param>
+        /// <return>A representation of the deferred computation of this call.</return>
+        Task PowerOffAsync(bool skipShutdown = false, CancellationToken cancellationToken = default(CancellationToken));
```
OrpheusQu: Same as above, don't give skipShutdown a default value.

src/ResourceManagement/Compute/Domain/IVirtualMachine.cs
weidongxu-microsoft: Yes. Remove the default value to skipShutdown. It will cause ambiguous method call.

PR #1039, opened 2020-04-09
Release 1.33.0, with breaking change

xccc-msft:


ChenTanyi:
we don't need to release .NET as we talk in the last meeting.

weidongxu-microsoft:
Nick need it since there is some pr that our guy needs.

ChenTanyi:
OK, got it.

qub1n:
In my opinion TaskCancellation handling is important breaking change and should be noted in release notes.

xccc-msft:
@qub1n Actually this is a fix/enhancement about better support for .NET Core/Framework. You may find the details via [this PR](https://github.com/Azure/azure-libraries-for-net/pull/1019#issue-391420832).

PR #1041, opened 2020-04-09
Add certificate roll over

### LINKED ISSUE #615: Support subject + issuer validation in AzureCredentialsFactory.FromServicePrincipal
In order to use a certificate that is whitelisted by subject + issuer instead of thumbprint, the whole public key needs to be sent when getting an access token. This is controlled by the sendx5c parameter in AuthenticationContext.AcquireTokenAsync. Currently it's not possible to set that flag through the AzureCredentialsFactory methods, and S+I validation doesn't work as a consequence.

A possible solution would be to add an optional sendx5c parameter to AzureCredentialsFactory.FromServicePrincipal, and pass it through to AcquireTokenAsync.

This may depend on the following issue, if that SDK is used internally for authentication rather than AuthenticationContext directly: https://github.com/Azure/azure-sdk-for-net/issues/5206

### LINKED ISSUE #1014: Not able to get token for Apps where certificates are whitelisted using issuer and subject name
AAD mandates that if we are using app where certificates are whitelisted using issuer and subject name; we should pass sendX5c flag to true.
In AzureCredential, when we call  'ApplicationTokenProvider.LoginSilentAsync', we don't have a way to pass IsCertificateRollOverEnabled to acquire token.

ChenTanyi:
There is signing problem in CertificateAuthenticationProvider. Fix first inside our project, and I'll open a PR at the azure-sdk-for-net
fix #615, fix #1014

### DIFF
--- a/src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
+++ b/src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
@@ -9,6 +9,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Security.Cryptography.X509Certificates;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


weidongxu-microsoft:
LGTM. I assume that this PR depends on the sdk fix?

Any detail of tests?

ChenTanyi:
@weidongxu-microsoft No, this PR is independent. The SDK PR is used for more easy to add  `IdentityModel.Clients.ActiveDirectory.ClientAssertionCertificate`. I currently add it in this PR.

ChenTanyi:
https://github.com/Azure/azure-libraries-for-net/pull/1041/files#diff-26cb4538b95188372407965ed5ce8a7cR238 Here

src/ResourceManagement/ResourceManager/Authentication/AzureCredentials.cs
```diff
@@ -175,8 +176,25 @@ public async override Task ProcessHttpRequestAsync(HttpRequestMessage request, C
 #else
                     else if (servicePrincipalLoginInformation.X509Certificate != null)
                     {
+                        IdentityModel.Clients.ActiveDirectory.ClientAssertionCertificate clientAssertionCertificate = 
+                            new IdentityModel.Clients.ActiveDirectory.ClientAssertionCertificate(servicePrincipalLoginInformation.ClientId, servicePrincipalLoginInformation.X509Certificate);
+                        CertificateAuthenticationProvider certificateAuthenticationProvider = 
+                            new CertificateAuthenticationProvider(clientAssertionCertificate, servicePrincipalLoginInformation.IsCertificateRollOverEnabled);
+
+                        credentialsCache[adSettings.TokenAudience] = await ApplicationTokenProvider.LoginSilentAsync(
```
yungezz, line 184: not sure the logic, why `certificateAuthenticationProvider` is a must now with CertRollOverEnabled

ChenTanyi: It was what the issue needed. The default `false` is not breaking with current logic.

ChenTanyi:
Sure, we could open it until next release. But it seems no any reply at sdk side in the PR.

Issue #1049, opened 2020-04-23
[QUERY]: Looking to create ScheduledQueryRules for log alerts


kazman71:
Looking to programmatically add Kusto queries for scheduled log alerts and I can't figure out how this might be done.  I can see the AlertRules.Manager.Inner.ScheduledQueryRules object, but I can't seem to understand how to specify the query.  Any help would be appreciated.  

Thanks in advance,
James

yungezz:
hi @weidongxu-microsoft do you have a quick answer? thanks

weidongxu-microsoft:
@kazman71 

It is not supported as Fluent API.

You can invoke it directly by
```
azure.MetricDefinitions.Manager.Inner.ScheduledQueryRules.CreateOrUpdateAsync
```

kazman71:
Thank you very much for the quick response - I will give it a try!

kazman71:
@weidongxu-microsoft 

Is there any example code using the LogSearchRuleResourceInner object with this API?  I seem to be getting a serialization error during the CreateOrUpdateAsync based on the contract.   See the images below.  Could it be the period in the property names?


![image](https://user-images.githubusercontent.com/6991201/80248825-8adef180-8625-11ea-9359-70b2298903b2.png)

![image](https://user-images.githubusercontent.com/6991201/80248781-769af480-8625-11ea-96df-2af3350fb047.png)

weidongxu-microsoft:
@kazman71 

The "dot" in the property name should be fine. It only means e.g. "properties.description" should be serialized to nested "description" under "properties", as in the sample.

Could you include the code of your configure and error stack?

Thanks.

kazman71:
Thanks for the response.  I seem to have narrowed it down to the Action parameter of LogSearchRuleResourceInner .  When I create a new AlertingAction object and  make the CreateOrUpdateAsynCall it will fail with the message: "Error reading string. Unexpected token: StartObject. Path 'error.message', line 4, position 16."

If I query for an existing rule and use the action object returned from an already created rule the createorupdate will succeed.  Thanks for looking into it.  When I serialize the the the input object to json it is seems the same as when I query for an already existing alert's schema. 

Here is the code to create the log alert:
            var locationString = "centralus";    //// TBD

            var appInsightsResourceId = await GetValue(deployment.Variables.Environment, "applicationInsightsResourceId", logger);
            var kustoQuery = await GetValue<string>(deployment.Variables.Component, "alertQuery", logger);
[...]

PR #1054, opened 2020-04-28
ignore error in ProviderRegistrationDelegatingHandler

weidongxu-microsoft:
Fix incorrect exception discovered from https://github.com/Azure/azure-libraries-for-net/issues/1049


weidongxu-microsoft:
Root cause is that service error response json is not on correct schema.

But anyway ProviderRegistrationDelegatingHandler should not throw on this. It is not a much used feature, and at best it is just a NOOP.

xccc-msft:
Should we print some warning message when catching this JsonException?

weidongxu-microsoft:
Generally there is not much tracing in SDK at all.

If the content is a exception, user will have it as an exception later in usual response handling, but not from here.

Issue #1056, opened 2020-04-30
NextPageLink property null although additional items exist


ya-alexander:
Hi,

I am trying to get all groups that exist in an AAD using the following code: 

`var groups = await this._azure.AccessManagement.ActiveDirectoryGroups.ListAsync();`

This call always returns exactly 100 items, although > 200 groups exist in AAD.  So I tried to query the rest of groups using the following code. 

```
var nextPage = await groups.GetNextPageAsync();
do
{
     allGroups.AddRange(nextPage);
} while (nextPage.Any());
```

Unfortunately, the NextPageLink property of `groups` is always null. However, if I trace the call of `this._azure.AccessManagement.ActiveDirectoryGroups.ListAsync()` in Fiddler, the `odata.nextLink` seems to be set correctly:

```
{
    "odata.metadata": "https://graph.windows.net/tenant-guid/$metadata#directoryObjects",
    "odata.nextLink": "directoryObjects/$/Microsoft.DirectoryServices.Group?$skiptoken=X'445370740200010000002A47726F75705F36366364373630312D386537302D343032652D613738612D3933633132393836386538302A47726F75705F36366364373630312D386537302D343032652D613738612D3933633132393836386538300000000000000000000000'",
    "value": [...]
}
```

I also tried to fetch additional groups in Fiddler using the `skiptoken` in a hand-crafted request and everything works as expected in Fiddler.

However, I cannot get to work pagination in my .NET code. I would expect that the `NextPageLink` property is != null if there are more than 100 groups available in AAD. 

Environment:
 - OS: Win10
 - IDE : VS 2019
 - Version of the Library: 1.33.0, but also tried various older versions 

Am I missing something, or could it be that pagination is not working correctly? 

Thanks in advance,
Alex

- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

allenjzhang:
Looking at the fiddler trace, the nextLink is NOT a valid full path URL. That may be the reason mgmt SDK is unable to retrieve the next page of data.

ya-alexander:
@allenjzhang Thanks for pointing that out. Is there a way I can influence the nextLink to be a full path URL? Am I using the library wrongly, or is this an issue of the underlying service?

allenjzhang:
Underlying service sets the return value. Mm, actually looked at the odata spec, the nextLink may be relative path to the current request replacing the last segment. Some services choose to return the absolute path but some may return relative path.

[...]

PR #1059, opened 2020-05-07
regen graphrbac to fix odata.nextLink

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/1056


src/ResourceManagement/Graph.RBAC/Generated/Models/Page.cs
```diff
@@ -24,7 +24,7 @@ public class Page<T> : IPage<T>
          /// <summary>
          /// Gets the link to the next page.
          /// </summary>
-         [JsonProperty("nextLink")]
+         [JsonProperty("odata.nextLink")]
```
weidongxu-microsoft, line 27: Seems graphrbac is not generated for long (or the swagger get fixed in meantime).

Issue #1062, opened 2020-05-13
[BUG] Diagnostic settings throwing NullReferenceException when dealing with subscription diagnostic settings


ni-reis:
**Bug description**
I'm trying to use IAzure.DiagnosticSettings to define and create a diagnostic setting for a subscription resource (`/subscriptions/{subscriptionId}`). The current implementation reaches the HTTP request code; the request succeeds and the diagnostic setting gets created, however, when parsing the response, it fails to read the metrics (as they are null) and throws a NullReferenceException.

**Exception message and stack**
{"Object reference not set to an instance of an object."}
at Microsoft.Azure.Management.Monitor.Fluent.Models.DiagnosticSettingImpl.SetInner(DiagnosticSettingsResourceInner inner) in C:\\Users\\reisn\\work\\azure-libraries-for-net-master\\src\\ResourceManagement\\Monitor\\DiagnosticSettingImpl.cs:line 139

**To Reproduce**
You can authenticate to fluent client with a service principal and invoke the method to create a diagnostic setting for the subscription resource. Here is a script that does it, streaming logs to an event hub:

```
using Microsoft.Azure.Management.Fluent;
using Microsoft.Azure.Management.ResourceManager.Fluent;
using System.Threading.Tasks;

namespace TestProject
{
    class Program
    {
        private const string ClientId = "<CLIENT ID>";
        private const string ClientSecret = "<CLIENT SECRET>";
        private const string TenantId = "<TENANT ID>";
        private const string SubscriptionId = "<SUBSCRIPTION ID>";
        private const string SubscriptionResourceId = "/subscriptions/" + SubscriptionId;

        private const string AuthorizationRuleId = "<AUTHORIZATION RULE ID>";
        private const string EventHubName = "<EVENT HUB NAME>";

        static async Task MainAsync()
        {
            var credentials = SdkContext.AzureCredentialsFactory.FromServicePrincipal(
                ClientId, ClientSecret, TenantId, AzureEnvironment.AzureGlobalCloud);

            var fluentClient = Azure
                .Authenticate(credentials)
                .WithSubscription(SubscriptionId);

            await fluentClient.DiagnosticSettings
                .Define("myDiagnosticSetting")
                .WithResource(SubscriptionResourceId)
                .WithEventHub(AuthorizationRuleId, EventHubName)
                .WithLog("Administrative", 1)
                .CreateAsync();
        }

        static void Main(string[] args)
        {
            MainAsync().Wait();
        }
[...]

PR #1066, opened 2020-05-14
Adding support for subscription diagnostic settings

ni-reis:
PR associated to issue https://github.com/Azure/azure-libraries-for-net/issues/1062 .

### DIFF
--- a/src/ResourceManagement/Monitor/DiagnosticSettingImpl.cs
+++ b/src/ResourceManagement/Monitor/DiagnosticSettingImpl.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Azure.Management.Monitor.Fluent.Models
         IDefinition,
         IUpdate
     {
+        internal static string SubscriptionDiagnosticSettingsUri = "/providers/AzureResourceManager/diagnosticSettings/";
         internal static string DiagnosticSettingsUri = "/providers/microsoft.insights/diagnosticSettings/";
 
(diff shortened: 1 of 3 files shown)


src/ResourceManagement/Monitor/DiagnosticSettingImpl.cs
```diff
@@ -24,6 +24,7 @@ internal partial class DiagnosticSettingImpl :
         IDefinition,
         IUpdate
     {
+        internal static string SubscriptionDiagnosticSettingsUri = "/providers/AzureResourceManager/diagnosticSettings/";
```
xccc-msft, line 27: [Subscription Diagnostic Settings](https://docs.microsoft.com/en-us/rest/api/monitor/subscriptiondiagnosticsettings/createorupdate) is different from [Diagnostic Settings](https://docs.microsoft.com/en-us/rest/api/monitor/diagnosticsettings/createorupdate#diagnosticsettingsresource). The `Subscription Diagnostic Settings` is not supported as per [package-2019-06](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/monitor/resource-manager#tag-package-2019-06), which is the current version we use.

ni-reis: I see, so this is not the right way to solve the issue?

weidongxu-microsoft:
feature supported via https://github.com/Azure/azure-libraries-for-net/pull/1071

PR #1108, opened 2020-07-16
Add methods to specify consumption pricing tier

### LINKED ISSUE #1107: [FEATURE REQ] Provide a more obvious way to create a consumption app service plan
**Is your feature request related to a problem? Please describe.**
When creating an app service plan, there's no obvious way to choose the "Consumption" pricing tier. The options are:
- `WithFreePricingTier`
- `WithSharedPricingTier`
- `WithPricingTier`, which accepts a `PricingTier`, but `PricingTier` doesn't have a predefined value for Consumption

**Describe the solution you'd like**
There should be a `WithConsumptionPricingTier` method.

**Describe alternatives you've considered**
As a workaround, it's possible to use this:
```csharp
.WithPricingTier(new PricingTier("Dynamic", "Y1"))
```

**Additional context**
N/A

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

thomaslevesque:
Fixes #1107

### DIFF
--- a/src/ResourceManagement/AppService/AppServiceBaseImpl.cs
+++ b/src/ResourceManagement/AppService/AppServiceBaseImpl.cs
@@ -185,6 +185,11 @@ namespace Microsoft.Azure.Management.AppService.Fluent
             return WithPricingTier(new PricingTier("Free", "F1"));
         }
 
+        public FluentImplT WithConsumptionPricingTier()
+        {
+            return WithPricingTier(PricingTier.ConsumptionY1);
+        }
+
         ///GENMHASH:256905D5B839C64BFE9830503CB5607B:7AC64BDE9A6045728A97AD3B7E256F87
         internal async override Task<SiteConfigResourceInner> GetConfigInnerAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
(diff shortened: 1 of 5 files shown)


src/ResourceManagement/AppService/AppServiceBaseImpl.cs
```diff
@@ -185,6 +185,11 @@ public FluentImplT WithFreePricingTier()
             return WithPricingTier(new PricingTier("Free", "F1"));
         }
 
+        public FluentImplT WithConsumptionPricingTier()
```
thomaslevesque, line 188: Not sure where the `///GENMHASH` on other methods come from... Is it supposed to be generated automatically?

weidongxu-microsoft: HASH is not needed.

In the implementation of this method, it should be able to use `ConsumptionY1` directly?

thomaslevesque: > In the implementation of this method, it should be able to use `ConsumptionY1` directly?

Yes, I thought about it, but I saw that the other methods didn't use the predefined pricing tiers, so I didn't either, for consistency. I can change it if you like.

weidongxu-microsoft: Yes, please.

thomaslevesque: Done (I amended the commit)

Issue #1113, opened 2020-07-22
[BUG] ListByResourceGroupAsync() making config/web call for every resource in the resource group resulting hitting the 12K read quota limit unlike it's sync counter part which make single call


rashanke2:
**Describe the bug**
We are calling the API ListByResourceGroupAsync() to fetch all the resources from the specific group. We have confirmed using Fiddler that it’s making separate config call for every request in the resource group. Where as ListByResourceGroup(), it's sync counterpart of the same API, unlike the async one, it makes singe request to pull all configurations.

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**

***Code Snippet***
var apps = await azure.AppServices.FunctionApps.ListByResourceGroupAsync(resourceGroup).ConfigureAwait(false);

**Expected behavior**
It should make a single config call. Making multiple config call is causing hitting the throttling limit. Ideally the behavior should be same in sync/async API.

**Screenshots**
![image](https://user-images.githubusercontent.com/37538221/88217596-6b834f00-cc7c-11ea-88c4-0c5fdbf4695c.png)
![image](https://user-images.githubusercontent.com/37538221/88226299-785a6f80-cc89-11ea-8cf2-d51041654e7d.png)

**Setup (please complete the following information):**
 - OS: Windows
 - IDE : VS 2017
 - Version of the Library used
Library and API Details
-	Nugget -> Assembly Microsoft.Azure.Management.ResourceManager.Fluent **version - 1.34.0**
-	Namespace used to read resource details from resource group -> Microsoft.Azure.Management.ResourceManager.Fluent.Core.CollectionActions
-	Namespace used for Function App operations(start/stop, etc.) -> Microsoft.Azure.Management.AppService.Fluent
-	API available in Microsoft.Azure.Management.ResourceManager.Fluent.Core.CollectionActions namespace
o	ListByResourceGroup
o	ListByResourceGroupAsync – Currently used API, which is firing too many get requests.

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ y] Bug Description Added
- [ y] Repro Steps Added
- [ y] Setup information Added

weidongxu-microsoft:
This is likely a bug on sync method. The design is bad (to send additional per resource requests), but that is the one there. For backward-compatibility I probably not able to do much.

[...]

PR #1116, opened 2020-07-27
improvement on list/listAsync of webapp/functionapp

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/1113

1. sync/async should behave same
2. document that list would make extra calls
3. devise lite version of the list to only return basic info of webapp/functionapp, provide a e.g. refresh to get full info

still in developing


weidongxu-microsoft:
Try to add a `IWebAppSimple` for simple list.

weidongxu-microsoft:
@ChenTanyi @xccc-msft @yungezz please take a look whether this make sense.

Previous `List` send 1 list, then n*2 get, which is not a good design.
Currently add a `ListBasic` that only send 1 list, and get basic info. If detail is needed, do a separate `GetById`.

For Java track2, plan is to use this `ListBasic` as the standard `List`.

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/1120

Include kind=api in list result for WebApp.

PR #1117, opened 2020-07-28
Regen cdn

weidongxu-microsoft:
Lots of generated code was not correct.

Fix https://github.com/Azure/azure-sdk-for-net/issues/13714


src/ResourceManagement/Cdn/Generated/EdgeNodesOperations.cs
```diff
@@ -91,7 +91,7 @@ public async Task<AzureOperationResponse<IPage<EdgeNodeInner>>> ListWithHttpMess
             }
             // Construct URL
             var _baseUrl = Client.BaseUri.AbsoluteUri;
-            var _url = new System.Uri(new System.Uri(_baseUrl + (_baseUrl.EndsWith("/") ? "" : "/")), "providers/Microsoft.Cdn/EdgeNodeInners").ToString();
+            var _url = new System.Uri(new System.Uri(_baseUrl + (_baseUrl.EndsWith("/") ? "" : "/")), "providers/Microsoft.Cdn/edgenodes").ToString();
```
weidongxu-microsoft, line 94: URL is just wrong...

weidongxu-microsoft:
> Do you have any idea if the issue was from swagger or generator?

No idea. I guess it is caused by a buggy out-dated csharp generator, and we did not regen it since took over the repo.

PR #1151, opened 2020-10-15
Compute: upgrade to package-2020-06-30

xccc-msft:


ChenTanyi (approved):

src/ResourceManagement/Compute/Generated/Models/DataDisk.cs
```diff
@@ -174,7 +174,7 @@ public DataDisk(int lun, DiskCreateOptionTypes createOption, string name = defau
         /// Gets or sets the managed disk parameters.
         /// </summary>
         [JsonProperty(PropertyName = "managedDisk")]
-        public Management.ResourceManager.Fluent.SubResource ManagedDisk { get; set; }
+        public ManagedDiskParametersInner ManagedDisk { get; set; }
```
weidongxu-microsoft: What causes this change? What's in swagger?

weidongxu-microsoft (approved):

src/ResourceManagement/Compute/Generated/Models/DataDisk.cs
xccc-msft: I think it's generator issue, already existing when we did upgrade last time [here](https://github.com/Azure/azure-libraries-for-net/pull/911/commits/bc14e74928c0e67c97b474da4cfafc2d87aafaf3)

Samples/Compute/ConvertVirtualMachineToManagedDisks.cs
```diff
@@ -52,7 +52,7 @@ public static void RunSample(IAzure azure)
                             .WithNewVhd(50)
                             .WithLun(2)
                             .Attach()
-                        .WithSize(VirtualMachineSizeTypes.StandardD3V2)
+                        .WithSize(VirtualMachineSizeTypes.Parse("Standard_D2a_v4"))
```
weidongxu-microsoft, line 55: What's the cause?

Samples/BatchAI/ManageBatchAI.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.Azure.Management.Storage.Fluent.Models;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.File;
+using ExecutionState = Microsoft.Azure.Management.BatchAI.Fluent.Models.ExecutionState;
```
weidongxu-microsoft, line 18: Why this change?

xccc-msft: Compute have same named `ExecutionState` after generate new version.

Samples/Compute/ConvertVirtualMachineToManagedDisks.cs
xccc-msft: For test purpose, we want to move vm size into AMD based.

PR #1152, opened 2020-10-18
Fixed the get method of Rules in AzureFirewallNatRuleCollectionImpl

eitanmechtinger:
Fixed the get method to return the existing rules instead of an empty list


ChenTanyi (approved):
Thanks for your contribution!

weidongxu-microsoft (approved):

Issue #1147, opened 2020-10-08
[BUG] ResourceManagement Graph.RBAC Lookup Active Directory Group by Name Returns Incorrect Group


mikemassa84:
**Describe the bug**
When using the Microsoft.Azure.Management.Graph.RBAC.Fluent library to retrieve an active directory group by name, an incorrect group is returned. For example, if looking up a group named "Foo", the api may return a group named "Bar" instead. In our case, we are searching for a group named "R&D - PA" but a group named "SG Bomgar Admins" is returned instead.

***Exception or Stack Trace***
No exception or stack trace.

**To Reproduce**
1. Setup multiple active directory groups in Azure Active Directory (we have hundreds).
  a. Create a group named SG Bomgar Admins and a group named R&D - PA.
2. Setup a project using the Microsoft.Azure.Management.Graph.RBAC.Fluent NuGet package.
3. Use code similar to that in the Code Snippet section to lookup group by name.

***Code Snippet***
Add the code snippet that causes the issue.

This is the code we are using to lookup group by name.
```C#
    Azure.Configure().WithLogLevel(HttpLoggingDelegatingHandler.Level.Basic).Authenticate(creds);
    var adGroup = await _auth.ActiveDirectoryGroups.GetByNameAsync("R&D - PA").ConfigureAwait(false);
```
Looking at the method in [ActiveDirectoryGroupsImpl.cs](https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/Graph.RBAC/ActiveDirectoryGroupsImpl.cs), it seems to be returning the first group that is found without verifying the name of the group matches the name requested:

```C#
IEnumerable<ADGroupInner> inners = await Inner.ListAsync(string.Format("displayName eq '{0}'", name), cancellationToken);
if (inners == null || !inners.Any())
{
    return null;
}
return WrapModel(inners.First());
```

**Expected behavior**
Based on our code above, we expect the adGroup object to contain a group named R&D - PA.

**Screenshots**
n/a

**Setup (please complete the following information):**
 - OS: Windows 10 (Also Windows Azure Functions)
 - IDE : Visual Studio 2019
 - Version of the Library used: 1.34.0

**Additional context**
We are able to work around the problem by loading all of the AD groups and looking up the group ourselves, however, loading all the groups is extremely slow. It takes between 35-40 seconds to complete. Once loaded though, the lookup is extremely fast.
```C#
[...]

PR #1154, opened 2020-10-20
GraphRbac: fix odata query filter with special characters

xccc-msft:
Fix: #1147


weidongxu-microsoft (approved):

ChenTanyi (approved):

ChenTanyi:
Maybe other `GetByNameAsync` should also be changed.

PR #1167, opened 2020-11-17
Fix regex to allow wildcard subdomain

SvyatovetsD:
I couldn't add hostname binding with wildcard subdomain though it can be done via azure portal. Tweaked regex expression a bit, so it could match wildcard

### DIFF
--- a/src/ResourceManagement/AppService/HostNameBindingImpl.cs
+++ b/src/ResourceManagement/AppService/HostNameBindingImpl.cs
@@ -140,7 +140,7 @@ namespace Microsoft.Azure.Management.AppService.Fluent
         ///GENMHASH:027705B70337BE533ED77421800E496A:3479885A232C974264B5B36BDBB0BC94
         public HostNameBindingImpl<FluentT, FluentImplT, DefAfterRegionT, DefAfterGroupT, UpdateT> WithDnsRecordType(CustomHostNameDnsRecordType hostNameDnsRecordType)
         {
-            var regex = new Regex("([.\\w-]+)\\.([\\w-]+\\.\\w+)");
+            var regex = new Regex("([.\\w-]+|[*])\\.([\\w-]+\\.\\w+)");
             var matcher = regex.Match(name);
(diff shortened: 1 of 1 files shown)


src/ResourceManagement/AppService/HostNameBindingImpl.cs
```diff
@@ -140,7 +140,7 @@ private string NormalizeHostNameBindingName(string hostname, string domainName)
         ///GENMHASH:027705B70337BE533ED77421800E496A:3479885A232C974264B5B36BDBB0BC94
         public HostNameBindingImpl<FluentT, FluentImplT, DefAfterRegionT, DefAfterGroupT, UpdateT> WithDnsRecordType(CustomHostNameDnsRecordType hostNameDnsRecordType)
         {
-            var regex = new Regex("([.\\w-]+)\\.([\\w-]+\\.\\w+)");
+            var regex = new Regex("([.\\w-]+|[*])\\.([\\w-]+\\.\\w+)");
```
ChenTanyi, line 143: It is better not to add `[` and `]`.

weidongxu-microsoft:
This seems not allowed on portal.

image

ChenTanyi (commented):
Hi @SvyatovetsD, we cannot use bind a web app with wildcard subdomain in Azure Portal, would you like to show a screenshot, gif or docs for it?

src/ResourceManagement/AppService/HostNameBindingImpl.cs
SvyatovetsD: It seems to work fine in my environment. I guess you got an error cause dns zone haven't been setup accordingly
Screenshot 2020-11-18 141755

weidongxu-microsoft: @SvyatovetsD
Could you see if you can apply @ChenTanyi 's advice above?

weidongxu-microsoft (approved):

PR #1203, opened 2021-02-02
Storage: manual revert on SkuName

xccc-msft:
To align: #901 
The storage service is not allowed to create using `Standard_RAGZRS` while it will return this if the storage account is created by other tools with newer API version.

### DIFF
--- a/src/ResourceManagement/Storage/Generated/Models/SkuName.cs
+++ b/src/ResourceManagement/Storage/Generated/Models/SkuName.cs
@@ -8,75 +8,25 @@
 
 namespace Microsoft.Azure.Management.Storage.Fluent.Models
 {
-    using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft (approved):
Do we have "RA-GZRS"? [10:44 AM] Vikranth Thati asked for that.

xccc-msft:
> Do we have "RA-GZRS"? Vikranth Thati asked for that.
> 
> If we can use it here, we might not need to upgrade to 2019-06

No. If we create using `RA-GZRS`, the service makes validation and returns error saying it's not allowed in current version. But if we get storage account with `RA-GZRS`, the service will not validate the value and returns the valid response.

weidongxu-microsoft:
@xccc-msft Do you need to add `Standard_RAGZRS` to the list? It might be in server response as well.

Issue #822, opened 2019-10-08
[BUG] Deployment Outputs object is not set on deployments created with Fluent API.


AlexeyEvlampiev:
**Description**
Deployment Outputs object is not set on deployments created with Fluent API.

**To Reproduce**
1. Create an ARM template producing some basic outputs, e.g.
`{
  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "resources": [],
  "outputs": {
    "result": { "type": "string", "value": "Ok" }
  }
}`
2. Initialize an azure object of the Microsoft.Azure.Management.Fluent.IAzure type, as described in [https://docs.microsoft.com/en-us/azure/virtual-machines/windows/csharp-template](url)
3. Deploy the above template (templateJson- parameter) to an existing resource group ( resourceGroup parameter) with 
`var deployment = azure.Deployments.Define($"myDeployment")
                .WithExistingResourceGroup(resourceGroup)
                .WithTemplate(templateJson)
                .WithParameters("{}")                .WithMode(Microsoft.Azure.Management.ResourceManager.Fluent.Models.DeploymentMode.Incremental).Create(); 
Debug.Assert(deployment.Outputs != null); `

***Code Snippet***
Debug.Assert(deployment.Outputs != null);

**Expected behavior**
The Debug.Assert(deployment.Outputs != null) assertion should pass.

**Setup:**
 - OS: Windows
 - IDE : Visual Studio
 - Microsoft.Azure.Management.Fluent 1.26.1

alex-frankel:
Hi Alexey,

I'm working on finding the team that manages this particular set of APIs to get more details on the fix we need to make. We manage the core .NET SDK, but not the Fluent APIs. Hoping to have an update to you tomorrow or Monday. Thanks in advance for your patience!

Alex

yaohaizh:
@AlexeyEvlampiev this is an issue of fluent SDK. 

For now, you can query the object via 
```csharp
                    deployment = resourceManager.Deployments.GetByResourceGroup(rgName, deploymentName1);
```
to populate the result.

yaohaizh:
Fix: https://github.com/Azure/azure-libraries-for-net/pull/830

AlexeyEvlampiev:
Thanks, @yaohaizh. The outputs object is indeed populated this way.

bwfilms:
[...]

PR #1230, opened 2021-04-02
bug fix, SetInner in async method

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-net/issues/822


ChenTanyi (approved):

src/ResourceManagement/ResourceManager/Deployment/DeploymentImpl.cs
```diff
@@ -423,17 +423,7 @@ public async Task<IDeployment> BeginCreateAsync(CancellationToken cancellationTo
 
         #region Implementation of ICreatable interface 
 
-        public new IDeployment Create()
```
weidongxu-microsoft, line 426: no idea why previously it need to use the `new` keyword...

PR #1262, opened 2021-08-27
Update AzureUSGovernment Graph endpoint

greganderson:
The graph endpoint is incorrect according to the Microsoft Graph: Developer Blog.  https://developer.microsoft.com/en-us/graph/blogs/new-microsoft-graph-endpoints-in-us-government-cloud/


weidongxu-microsoft:
@greganderson 

This is the legacy Active Directory Graph endpoint, not the MS Graph.

I think this is the new MSGraph client https://github.com/microsoftgraph/msgraph-sdk-dotnet

PR #1272, opened 2021-09-23
configuration as release

weidongxu-microsoft:


.azure-pipelines/fluentNetSdk.yml
```diff
@@ -46,11 +46,13 @@ jobs:
         displayName: 'Running tests'
         inputs:
           solution: build.proj
+          configuration: release
```
ChenTanyi, line 49: Does L44 need it?

ChenTanyi (approved):

.azure-pipelines/fluentNetSdk.yml
weidongxu-microsoft: Not sure, will add it.

